PR #7, opened 2015-03-12
This fixes 8 broken test situations. 

pilz:
Reasons were stale tests which were never run before but contained code that is irrelevant or wrong in the PI context. Also we have conflicting profile imports.

@mattss @wengole @gyst I am a bit clueless about the rest of the failures.


mattss:
Superceded by #8 - we have now ripped out slc.\* eggs from ploneintranet.attachments

PR #44, opened 2015-03-13
Bootstrap: use the pip based bootstrap.py

cillianderoiste:


cillianderoiste:
@gyst, does this look OK to you?

PR #46, opened 2015-03-13
Handover tags

pysailor:
Hand over selected tags, so that if the user has already selected tag(s) and then does a search, the new tag does not replace the existing ones, but gets added.

Since selecting tags and searching for tags is handled by two separate forms with two separate actions, we need to play a little game of pat-inject ping-pong.

Also, here we remove the unwantend "Select All / None" selector in the tags panel


pysailor:
@mattss or @wengole or @gyst Could you please review this and merge if OK?
I'll be happy to elaborate on the pat-inject ping-pong if you want.

mattss:
I will hopefully have a moment to look through this at some point this week. At the moment it looks like it needs another master merge(!)

src/ploneintranet/activitystream/browser/activity_provider.py
```diff
@@ -135,7 +136,8 @@ def Creator(self):
     @property
     def text(self):
         url = api.portal.get().absolute_url()
-        text = link_users(self.context.text, url)
+        text = safe_unicode(self.context.text)
+        text = link_users(text, url)
```
pysailor, line 140: This might seem strange, but it's only an interim stage. With the changes in the fix-mentions branch (yet to be PRed and merged), creating the text to show in the status will be streamlined.

mattss:
Just tried this out and it seems to work well. thanks @pysailor !

PR #50, opened 2015-03-13
Remove code duplication for comment replies

pysailor:
Make code simpler, re-use the template & macro structure that we have for this purpose


mattss:
@pysailor Can you remind me what the comment-well-said template is for?

pysailor:
@mattss That is the template that is used as a reply when a new comment was posted on an existing status update. In proto this is https://github.com/ploneintranet/ploneintranet/blob/master/prototype/feedback/comment-well-said.html
As you can see here https://github.com/ploneintranet/ploneintranet/blob/master/prototype/_includes/post.html#L98 this is used as the action for injection.

This chapter in the documentation more or less explains it
https://github.com/ploneintranet/ploneintranet/blob/master/docs/development/plonesocial.rst#creating-a-post

mattss:
Wow! Great job on documentation!

pysailor:
thanks ;) Not 100% there yet, but basically a brain-dump

PR #59, opened 2015-03-14
Case task sidebar enhancements

### LINKED ISSUE #3: workspace sidebar should maintain colour
If you go direct to an item in a workspace you only get the default colour

### LINKED ISSUE #58: Sidebar tasks should be open when navigating to a task
This should happen if you go directly to a task in a workspace.

adrianwhite:
@cornae you may want to review these changes.
I added an extra class to the div.application so we can distinguish between a workspace type and page type. I updated the css accordingly: 04c03bd and f672b6b

Unfortunately I cannot sort #58 as this is an issue in pat-stacks

This pull fixes #3


cornae:
So that's the same ' type-case' and ' type-…' that I already added there, but now it appears on all pages within a workspace context and not only the workspace landing page, right?

adrianwhite:
I changed it as I could forsee where we would want to know the type of the context we are on not just the type of the workspace we are in.

PR #62, opened 2015-03-14
bootstrap_jenkins: use virtualenv --relocatable to reduce shabang length

cillianderoiste:
When jenkins jobs have long names the shabang path created by virtualenv
exceeds the maximum length allowed of 127 characters. We can work around
this by using `virtualenv --relocatable .` after any console scripts are
added to ./bin.


cillianderoiste:
erm, "shebang" not "shabang" :D

PR #82, opened 2015-03-16
sidebar open by default when viewing a document directly

### LINKED ISSUE #77: The sidebar should be open by default when viewing a document directly
This requires applying the sidebar-left-open class to the body.

pbauer:
fix #77

### DIFF
--- a/src/ploneintranet/theme/static/rules.xml
+++ b/src/ploneintranet/theme/static/rules.xml
@@ -66,8 +66,17 @@
   <replace css:content-children="#social-messages" css:theme="#document-content" />
 </rules>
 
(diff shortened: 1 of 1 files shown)


mattss:
The diazo rules are pretty horrific, and adding to them makes me sad.

pbauer:
@mattss @pilz any alternative solutions?

mattss:
Unless @pilz has any, I think that re-thinking the diazo rules is probably a separate, larger task unfortunately.

pbauer:
then we should merge this until we have a better solution

mattss:
Agreed :(

pilz:
Yes, but we will look at the diazo rules soon :) All will be good!

PR #86, opened 2015-03-17
Sidebar markup fixes

pilz:
Updated the markup in our browser views to match the recent changes to the prototype. This fixes the missing controls in the sidebar and updates the workspace tabs.


pysailor:
Looking good, well done!

PR #88, opened 2015-03-17
Correct name for template

pysailor:
Rename wrongly-named upload.html to update-social.html, and remove the note about this in the documentation.


mattss:
Good call.

PR #89, opened 2015-03-20
Definition of done

wengole:


docs/development/contributing.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Definition of Done
+==================
+
+In order for a pull request to be accepted
+the following development and style guidelines must be adhered to:
+
+ * The `continuous integration <http://jenkins.ploneintranet.org>`_ build must be
+ passing
+ * Any contributed code, must have 100% test coverage
```
mattss, line 36: I think the 100% test coverage ship has sailed. It would be good to work out what coverage we have now, and set that as the target so we don't add any _more_ untested code.

pbauer, line 36: @mattss I think "Any contributed code must have 100% test coverage" says exactly that.

wengole, line 36: @pbauer I too agree also. :D

PR #90, opened 2015-03-20
Update the Subject index after setting tags

### LINKED ISSUE #70: Testing profile fails to create entries in Subject index
Create a new Plone site and apply the "Plone Intranet: Suite: Create Testing Content" profile.
Even though news items are created that also have Subjects (e.g. /news/alliance-officially-opens-new-lounge), the "Subject" index is empty.
Manually re-indexing one such news items does not cause the "Subject" index to be populated. Manually re-indexing the "Subject" index however works: Afterwards, the three test news items and their two Subject values are indexed.

mattss:
Resolves https://github.com/ploneintranet/ploneintranet/issues/70

### DIFF
--- a/src/ploneintranet/suite/setuphandlers.py
+++ b/src/ploneintranet/suite/setuphandlers.py
@@ -174,7 +174,7 @@ def create_news_items(newscontent):
         # api.content.transition(obj=obj, transition='publish')
 
         obj.setEffectiveDate(newsitem['publication_date'])
-        obj.reindexObject(idxs=['effective', ])
+        obj.reindexObject(idxs=['effective', 'Subject', ])
 
 
 def create_tasks(todos):


src/ploneintranet/suite/setuphandlers.py
```diff
@@ -174,7 +174,7 @@ def create_news_items(newscontent):
         # api.content.transition(obj=obj, transition='publish')
 
         obj.setEffectiveDate(newsitem['publication_date'])
-        obj.reindexObject(idxs=['effective', ])
+        obj.reindexObject(idxs=['effective', 'Subject', ])
```
pysailor, line 177: I had actually tried that too before I opened the ticket, but it did not work for me then. But now I tried it again, and all is fine. I must have overlooked something then.

PR #91, opened 2015-03-20
Fix naming issues with social packages

mattss:
- PloneIntranet -> Plone Intranet for consistency with existing
  packages
- Add 'Social' to some of the social eggs to remove confusion with
  existing eggs (e.g. theme)

### DIFF
--- a/src/ploneintranet/activitystream/configure.zcml
+++ b/src/ploneintranet/activitystream/configure.zcml
@@ -28,7 +28,7 @@
 
   <genericsetup:registerProfile
       name="default"
-      title="PloneIntranet Activitystream"
+      title="Plone Intranet Activitystream"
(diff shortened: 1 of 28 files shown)


mattss:
@gyst @pysailor Quite a few changes here but all text/translation related. Hopefully this will make things clearer!

pysailor:
Looks good to me. But @gyst you're the "owner" so I think you should have the final say & merge.

mattss:
@gyst do you want to retain final say on plone social changes like this? Or should we just get on with this kind of thing now that we've merged the codebases?

gyst:
@mattss This is really just infrastructure maintenance for which I don't need veto rights :-). Also I'm getting so much mail I might not notice a mention.

Issue #49, opened 2015-03-13
Create buildout.d directory


wengole:
There are quire a lot of buildout configs now. To keep the root of the project tidy, @cillianderoiste and I discussed having a buildout.d directory

pilz:
We have one now

wengole:
That's good then. 😀

Wondered why I was getting a whole load of ploneintranet emails, this issue was created by me!

pilz:
Hehe, you will get more, probably. Major cleanup ahead!

PR #92, opened 2015-03-22
Fix mentions

pysailor:
This enables mentioning other user in a Status update or reply to it.
Similar to tagging, mentions are now stored in a separate list and are no longer scraped from the text of the post. And like for tagging (#49), you can select one or more users in the drop down, search/filter for more users, go back to the full list and pick some more - the currently selected users are handed over.

I will add robot tests for mentioning (and tagging) in a separate branch / PR


mattss:
Just looking at this one now

PR #111, opened 2015-03-25
Workspace security pass 1

wengole:
Have hidden the workspace settings tab if user does not have permission to manage the workspace.

This is not the ideal solution, but as an interim it fullfils the issue where non-members of open workspace could (seemingly) edit the workspace settings. It transpires that the backend logic was sound, and these changes were not taking effect.

Pass 2 will involve returning 403s and disabling UI (sliders) for workspace settings, but as this runs the risk of not being done by the end of this sprint, this PR can be merged as an interim solution


wengole:
@gyst This can (with proper review by someone who's not me) be merged, or we can wait until I've finished off "pass 2" which I should get done by the end of this week.

pbauer:
:-1: on waiting

wengole:
Superseded by #126

gyst:
Ben, reviewing your PR I think 3da9ffa7cb3487 should be reverted. 

Checking the current user's view permission requires access to the REQUEST (which represents the user) and should consequently IMO be done in a browser view (which adapts REQUEST). The fact that the adaptation is hidden in plone.api doesn't change that. 

It seems to me an interface contract violation if a content object, which is REQUEST unaware, now implicitly checks the request.

But then I don't know the reason why you factored this method out of the tile to start with?

PR #115, opened 2015-03-25
experimental api

pbauer:
Do not merge!  This is a RFC. 

I believe we api-like convienience methods for functionality provided by ploneintranet. IMHO each python-package should have a `api.py` (or similar) that offers nice methods to make use of the most important features in said package. This is an example with only one method.


gyst:
Merged.

PR #116, opened 2015-03-25
104 code analysis

gyst:
Fix code-analysis; cleanup all warnings and errors; activate pre-commit hook.


wengole:
@gyst FYI in python, you can concatenate strings but surrounding them in parentheses. This is preferable to using backslash for line continuation. This also applies to imports. See [PEP 328](https://www.python.org/dev/peps/pep-0328/#rationale-for-parentheses)

PR #128, opened 2015-03-28
We return and display a link to the file and the file preview image

ale-rt:
Since the recent merge, the attachment returned by the provider is a dict.

Attaching a file to a comment will turn out in a broken preview.

It should contain two keys:
- link:  the link to the original file
- img_src: the link to the preview image

I am using "|attachment" as fall back for backward compatibility, i.e. to handle the case where a provider returns just a string with a link.


ale-rt:
How to test this:
1. Make a comment to a post and attach something
2. Reloading the page without this patch will result in a broken preview

Issue #127, opened 2015-03-27
stream file upload UTF8-safe


gyst:
This follows up on #57.

If you upload bärtige_flößer.odt instead of basic.txt you get a hanging screen. I already removed one bug in 0d1db5c but apparently more needs to be done.

ale-rt:
I cannot reproduce this anymore. Is it already fixed?

gyst:
Yes I fixed that ages ago. Haven't maintained the issue tracker.

## 

   Guido Stevens  |  +31.43.3618933  |  http://cosent.nl

   s o c i a l   k n o w l e d g e   t e c h n o l o g y

PR #129, opened 2015-03-28
Attempt to fix issue #127 (non-ascii in filenames on stream attachments)

pysailor:
This was a real bitch to track down....

The initial solution to get the upload working was pretty trivial: use Plone's standard IURLNormalizer to get a 'safe' id for the attachment.

The challenge was to find out that the same normalization needs to be applied also in pop_temporary_attachment() of ploneintranet.attachments.utils:
https://github.com/ploneintranet/ploneintranet/commit/7064738ecbf4b4da246cc44234152cf5c2d65b3c#diff-076a3598523e807f22b0c9945321bd57R42
But we're dealing with a different kind of object from which the id is generated:
In pi.attachments.upload we get the id from a p.a.contenttypes 'File' object, which is unicode: `u'Ba\u0308rtige Flo\u0308\xdfer.docx` which gets normalized to `bartige-flosser.docx` . In pop_temporary_attachment we get the id from a FileUpload instance, which is stored as utf-8 encoded string: `Ba\xcc\x88rtige Flo\xcc\x88\xc3\x9fer.docx` - normalize however makes `bairtige-floiaer.docx`out of it. Therefore this id needs to be turned into unicode again before we can apply the normalization. THEN the uploaded file is correctly found in the temporary attachment storage.

@gyst and @ale-rt This seems to fix all issues I had with non-ascii in the names of files and images. 2nd and 3rd pair of eyes much appreciated...
@reinhardt this also seems to fix the failing conversion we were looking at last week.

The test are still green...

### DIFF
--- a/src/ploneintranet/attachments/browser/upload.py
+++ b/src/ploneintranet/attachments/browser/upload.py
@@ -3,9 +3,11 @@ from plone import api
 from plone.memoize.view import memoize
 from plone.app.contenttypes.content import File
 from plone.app.contenttypes.content import Image
+from plone.i18n.normalizer.interfaces import IURLNormalizer
 from ploneintranet.attachments import utils
 from ploneintranet.attachments.attachments import IAttachmentStorage
 from ploneintranet.docconv.client.interfaces import IDocconv
+from zope.component import getUtility
 import logging
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


ale-rt:
I have looked at this and this seems fine to merge

gyst:
That was a real bitch to bring under test. Turns out the actual attachment conversion works fine and shows up in the stream OK, but the initial javascript injection has UTF-8 errors (not only with the filename, also with the username), but ONLY in the testrunner setup not in a real site. Tried to hunt that down but stopped wasting my time and skipped checking the injection.

gyst:
Confusing Github UX :-(

PR #140, opened 2015-05-04
Remove contentactivity provider

ale-rt:
Removed the from the stream all the content related activities


src/ploneintranet/activitystream/browser/stream_provider.py
```diff
@@ -74,9 +71,7 @@ def render(self):
     __call__ = render
 
     def activities(self):
-        brains = self._activities_brains()
```
gyst, line 77: This also starts to remove discussion. OK with me.

PR #145, opened 2015-05-04
Remove proto

jcbrand:
The prototype directory has been removed and is now [in its own repo](https://github.com/ploneintranet/ploneintranet.prototype).

The Makefile has been updated to do the following when running `make` or `make diazo`:
1. Check if the `prototype` dir exists and if not, do `git clone` of the prototype repo.
2. Run `make jekyll` in the prototype in order to build the HTML files from the jekyll source files.
3. Run `make fetchrelease` in ploneintranet to download the latest released javascript bundle
4. Copy the relevant files from the prototype to src/ploneintranet/theme/static/generated from where they are used by Diazo.

Those steps are roughly the same as before except for the cloning of the prototype repo.


gyst:
@jcbrand This breaks our jenkins so I'm holding off on the merge until that is fixed.

pysailor:
The reason seems to be a simple bug in https://github.com/ploneintranet/ploneintranet.prototype/blob/master/Makefile#L43, which I can't fix, because I don't seem to be a contributor on ploneintranet.prototype
@jcbrand (or @gyst  since you're an Owner in ploneintranet): Can you please enable access to all Contributors?

mattss:
@pysailor I've just added ploneintranet.prototype repo to the Contributors group.

PR #147, opened 2015-05-04
hide unsupported interface elements

giacomos:


src/ploneintranet/activitystream/browser/navigation.py
```diff
@@ -39,17 +39,20 @@ def items(self):
...
-        if PLONEINTRANET.network:
-            menu.extend([dict(url=base + '@@stream/network',
-                              title='My network',
-                              state='stream'),
-                         dict(url=base + '@@author',
-                              title='My profile',
-                              state='profile')])
+
+#        TODO: these lines are commented until the corresponding
+#        functionalities  will be complete
```
gyst: No these lines should be removed completely. The dash navbar should not be dynamic.

src/ploneintranet/workspace/browser/tiles/templates/sidebar.pt
```diff
@@ -72,6 +72,7 @@
                             <span class="float-before button-cluster select-buttons">
                                 <button type="button" class="select-all" title="Select all">All</button>
                                 <button type="button" class="deselect-all" title="Deslect all">None</button>
+                                <h1>22222</h1>
```
gyst: Why this extra h1?

src/ploneintranet/theme/static/rules.xml
```diff
@@ -38,10 +38,6 @@
   <replace css:theme-children="#home-tabs"
```
gyst: This one should be removed as well.

gyst:
Depends on PR #152 being merged first.

PR #149, opened 2015-05-05
4446 remove activitystream portal

jaroel:


ale-rt:
After this was merge we have test failing complaining in this way:

```
Failure in test Home view (functional/member.robot) #member #home
Traceback (most recent call last):
  File "/opt/ploneintranet/eggs/unittest2-0.5.1-py2.7.egg/unittest2/case.py", line 340, in run
    testMethod()
  File "/opt/ploneintranet/eggs/robotsuite-1.6.1-py2.7.egg/robotsuite/__init__.py", line 455, in runTest
    assert last_status == 'PASS', last_message
AssertionError: Page should have contained text 'The "Explore" section shows all updates of all people.' but did not
```

The tests are defined here:
- https://github.com/ploneintranet/ploneintranet/blob/251c8cf9f1e69c38030b6b6ac2f7c93c86ae1e60/src/ploneintranet/socialsuite/tests/functional/member.robot

It seems we are missing an import step from the demo profile:
- https://github.com/ploneintranet/ploneintranet/blob/680ad17041a268a59a84a83faa9ddf4253f56afd/src/ploneintranet/socialsuite/setuphandlers.py

PR #154, opened 2015-05-05
Remove not needed conditional imports

ale-rt:


src/ploneintranet/activitystream/browser/activity_provider.py
```diff
@@ -144,10 +136,10 @@ def text(self):
         return text
 
     def is_attachment_supported(self):
-        return IAttachmentStoragable is not None
+        return True
 
     def is_preview_supported(self):
-        return IDocconv is not None
+        return True
```
jaroel, line 143: Can we remove the checks for is_preview_supported and is_attachment_supported? They return True everywhere.

mattss, line 143: I think the point of these methods in plonesocial was that there were a variety of 'providers', some of which may or may not support previews and attachments. Having said that, I think that previews are pretty much a requirement for everything in ploneintranet, so this may indeed no longer be relevant. @gyst / @pysailor any thoughts there?

ale-rt, line 143: Yes!!! That is definetely the goal :)
I was setting up a task to address this (see wb 4460).
I wanted this pull request to be very specific in removing the ImportError checks.

PR #155, opened 2015-05-05
#4459 Clean up stream provider

jaroel:


src/ploneintranet/activitystream/browser/stream_provider.py
```diff
@@ -90,15 +88,12 @@ def activities(self):
                 i += 1
 
     def _activity_visible(self, activity):
-        if IStatusActivity.providedBy(activity) and self.show_microblog:
+        if IStatusActivity.providedBy(activity):
```
ale-rt: is enough to return IStatusActivity.providedBy(activity)

Probably it is better to remove completely this method and just substitute self._activity_visible with IStatusActivity.providedBy

```
[ale@kenobi ploneintranet]$grep -R _activity_visible src/ploneintranet
src/ploneintranet/activitystream/browser/stream_provider.py:            if self._activity_visible(activity):
src/ploneintranet/activitystream/browser/stream_provider.py:    def _activity_visible(self, activity):
```

src/ploneintranet/activitystream/browser/stream_provider.py
```diff
@@ -90,15 +88,12 @@ def activities(self):
...
             return True
         return False
 
     def _activities_statuses(self):
-        if not self.show_microblog:
-            raise StopIteration()
         container = PLONEINTRANET.microblog
```
ale-rt, line 91: I think also container is always True

gyst, line 91: NO! It's the outcome of queryUtility i.e. the actual microblog container tool.

ale-rt, line 91: Ok... here I can see this:
- https://github.com/ploneintranet/ploneintranet/blob/251c8cf9f1e69c38030b6b6ac2f7c93c86ae1e60/src/ploneintranet/core/integration.py#L25

So I think container is always Truish...

If it always evaluate to True, the "if not container" in the line below is redundant.

PR #157, opened 2015-05-05
Search api docs

mattss:
@mgrbyte and @adamcheasley here is our stab at the API interface/docs. Let us know if it makes sense and if we've missed anything obvious.


mattss:
Sorry this is the wrong place for this pull request.

PR #167, opened 2015-05-07
Remove activity provider

ale-rt:


.gitignore
```diff
@@ -20,7 +20,7 @@ lib/
 libreoffice/
 node_modules/
 parts/
-prototype
+/prototype
 release/
```
ale-rt, line 24: This is needed otherwise git will ignore all the prototype folders (not just the one where we checkout the ploneintranet.prototype)

src/ploneintranet/activitystream/adapters.py
```diff
@@ -10,6 +11,7 @@ class StatusActivity(object):
 
     def __init__(self, context):
         self.context = context
+        self.getId = context.getId()
```
ale-rt, line 14: This is needed to be fully compliant with the IActivity interface

gyst:
449 additions and 1,230 deletions. Incredible job @ale-rt !

PR #168, opened 2015-05-07
Add support for Wysiwyg editing with pat-raptor and better notification support.

jcbrand:
The follow features were added in this pull request:
- Wysiwyg editing with pat-raptor now works.
- Fixed a bug where the "Unsaved changes" popup would appear even though the page was saved.
- Status messages are now handled via the globalstatusmessage viewlet.
- The states shown in the workflow dropdown are now computed and not hardcoded.


jcbrand:
Thanks a lot @pysailor!

PR #169, opened 2015-05-07
Decouple prototype

gyst:
This completely decouples the prototype, meaning you can install ploneintranet without checking out the prototype. Two advantages:
- you don't need jekyll and the whole prototype build tooling
- changes in the prototype cannot "sneak into" ploneintranet without being tracked by a git commit

The 'make diazorelease' target updates ploneintranet based on a prototype checkout as follows:
- copies over only the html page templates referenced in rules.xml
- copies over all CSS stylesheets and assorted styling resources
- does NOT copy over javascript

Javascript is pulled in via HTTP. It is pinned by the LATEST revision pin, which gets automatically updated when running 'diazorelease'.

@jcbrand @pilz can you please review?


gyst:
When reviewing it's probably easiest if you ignore f5d42ad ...

gyst:
The github-jenkins integration is incomplete! I get a test failure on this branch on Jenkins - even though here it says "All is well - 13 successfull checks".

The test failure below is Jenkins specific.
http://jenkins.ploneintranet.net/view/All/job/complete.Ploneintranet.all.branches.even.broken.True/247/consoleFull#-987094442b55daa74-4590-49a8-959b-ec8d0928d875

If I locally run "git clean -fdx && make all devel test" all tests pass.

Makefile
```diff
@@ -1,59 +1,95 @@
...
 clean:
 	rm bin/* .installed.cfg || true
 check-clean:
 	test -z "$(shell git status --porcelain)" || (git status && echo && echo "Workdir not clean." && false) && echo "Workdir clean."
 
+fetchrelease:
+	# update LATEST in case we updated the prototype
+	$(eval LATEST := $(shell cat LATEST))
+	$(eval BUNDLEURL := https://products.syslab.com/packages/$(BUNDLENAME)/$(LATEST)/$(BUNDLENAME)-$(LATEST).tar.gz)
+	echo $(BUNDLEURL)
```
jcbrand, line 18: I'm now sure what you mean with `# update LATEST in case we updated the prototype`, looks to me like you're just reading the contents of LATEST, not checking what LATEST is in the prototype.

Also, looks to me like we would manually have to update the LATEST file in ploneintranet if we have done `make jsrelease` in the prototype, correct?

Makefile
```diff
@@ -1,59 +1,95 @@
...
+	# replace minfied js bundle with dev bundle, directly in diazo theme dir
+	cp prototype/bundles/$(BUNDLENAME)-dev.js $(DIAZO_DIR)/bundles/
+	sed -i -e 's#$(BUNDLENAME).min.js#$(BUNDLENAME)-dev.js#' $(DIAZO_DIR)/*.html
+
+bundle: prototype
+	cd prototype && make bundle
+
+jsrelease: prototype
+	cd prototype && make jsrelease
+
```
jcbrand, line 93: All in all the Makefile now looks really good. Clear and understandable.

Makefile
jcbrand, line 18: On that note, I think we need to add a step `jsrelease` in this Makefile, which delegates to `jsrelease` in the prototype Makefile and which will make sure that LATEST is up to date.

docs/installation/development.rst
```diff
@@ -104,6 +92,15 @@ Running tests
...
+--------------------
+
+The theme is derived from a separate package: ploneintranet.prototype.
+
+Documentation TODO.
+https://github.com/ploneintranet/ploneintranet.theme/blob/0ba4c21acd6da8e941fa94d0421491e59228009d/README.rst
+
+
 Building the Documentation
 --------------------------
```
jcbrand, line 106: I think we should also document the steps for doing Javascript development.

Looking at the Makefile, those steps appear to be:

``` bash
make jsdev # Checks out the prototype, makes a local javascript bundle, builds jekyll and copies HTML files to Diazo dir
# When ready to release JS changes:
# (perhaps rather a jsrelease step can be added to ploneintranet's Makefile which does all the below steps).
cd ./prototype && make jsrelease && cp LATEST ../  &&cd .. && git commit LATEST -m "Updated js release"
```

Makefile
gyst, line 18: ploneintranet has it's own LATEST which is updated automatically from the prototype in the 'make prototype' step. Because the Makefile variables $(LATEST) and $(BUNDLEURL) are defined at startup time (parsing the Makefile) we need to re-set those variables.

I.e. if you run 'make diazorelease' that also updates the prototype, including the LATEST stamp there, which means $(LATEST) and $(BUNDLEURL) have to be recalculated.

There is also a 'make jsrelease'.

docs/installation/development.rst
gyst, line 106: I will port the old theme documentation to also document prototype development.
Am thinking jsrelease belongs in the prototype though. You'd need to commit your changes there anyway first before making a release.

gyst:
This has now been merged directly.

PR #173, opened 2015-05-08
Fix workflow change

pysailor:


pbauer:
This is not ready to be merged.

Changing the workflow once works. On changing it again (without reloading the page) it tries to change the state twice (and the second attempt to to the same transition naturally fails):

```
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module ploneintranet.theme.content.content, line 29, in __call__
  Module ploneintranet.theme.content.content, line 42, in update
  Module plone.api.content, line 2, in transition
  Module plone.api.validation, line 70, in wrapped
  Module plone.api.content, line 2, in transition
  Module plone.api.validation, line 131, in wrapped
  Module plone.api.content, line 2, in transition
  Module plone.api.validation, line 100, in wrapped
  Module plone.api.content, line 379, in transition
InvalidParameterError: Invalid transition 'submit'.
Valid transitions are:
publish
send_back
```
- The number of popup-messages doubles each time i do a transaction (it gets reset to 1 after reloading the page.) All messages say that the sate has changes even though most transitions have failed (see above) 
- The submit-button is not long enough (you can see it in the case of the state "published". The active area is smaller than it should be.

gyst:
The issues @pbauer notes are design issues that need to be resolved in the prototype. I'll take this up with @cornae - see https://github.com/ploneintranet/ploneintranet.prototype/issues/32

pbauer:
@gyst does that mean we can still merge this pr?

gyst:
@pbauer No it means it's impossible to fix this without design changes. But if you think this is an improvement over master then it's OK to merge. I.e. if we get at least the initial transition working that's better than having now workflow transitions at all.

pbauer:
I do think it is a significant improvement over not being able to change the wf at all. So :+1: for merging.

gyst:
This has now been merged directly.

cornae:
The prototype has merely one example in of how the self healing banner is triggered after a workflow state with the idea that this would be enough specification to build it. I'm not sure if more simulation in the prototype is feasible, but is it necessary?

pysailor:
@cornae The notifications are fine. Any issues with them are backend-issues (we're still trying to hunt that one down).
The other issue (submit element not long enough) is a result of incorrectly copied markup.
So both issues that we need to fix in the backed

cornae:
Ok. Let me know if any prototype markup needs clarification, I'll be happy to help out.

PR #177, opened 2015-05-08
wrong TAL variable name

wengole:
Also testing new jenkins config. Don't merge until tested..


mattss:
OK Jenkins doesn't seem to be reporting status, but I've manually checked and it has passed all tests on this branch here: http://jenkins.ploneintranet.net/job/complete.Ploneintranet.all.branches.even.broken.True/259/

PR #197, opened 2015-05-12
Sidebar grouping

pilz:
This implements the grouping dropdown in the sidebar of a workspace. It implements a grouping storage for quick access to the different aggregations and adds a catalog index and metadatum for mimetype. This potentially clashes with mimetype indexes in the solr branch, so once this gets merged, we should look at the sidebar again manually.


do3cc:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/8/
Test PASSed.

src/ploneintranet/workspace/adapters.py
```diff
@@ -88,3 +98,270 @@ def getRoles(self, principal_id):
...
+                if uid in groupings[grouping][value]:
+                    continue
+                else:
+                    groupings[grouping][value].add(uid)
+                    groupings[grouping][value]._p_changed = 1
+            else:
+                groupings[grouping][value] = GroupingStorageValues([uid])
+
+        groupings[grouping]._p_changed = 1
+        groupings._p_changed = 1
```
gyst: I don't like the sight of these ._p_changed. It's brittle code and if you somehow miss one of the levels you're in big trouble. If you're using a Persistent storage object and you're mutating that object, you shouldn't then have to tell that object that it was mutated. That should be an internal of the storage object, not something you need to manage from the calling code.

In my own code I've used OOBTree.OOTreeSet and the like for similar storage purposes. 
https://pythonhosted.org/BTrees/

That gives you two benefits: 
1) no need to manually manage ._p_changed 
2) use the .insert() method to do away with the nested if/if/else/else state checks. Insert inserts a value if it is not there, but _does nothing_ if the value is already there, meaning you can safely use it to _always_ initialize a datastructure without risking existing data getting lost.

See e.g.
https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/network/graph.py#L27

gyst: PS the reason that you need to call _p_changed is because the innermost uid storage is a plain set. Replacing that with a truly persistent object should remove the need for manual _p_changed cascades.

gyst:
I have a patch coming up for OOTreeSet

gyst:
When testing interactively, there's an issue with pat-inject. If I change the grouping on a folder, I only get a 'loading' icon. When hard reloading the page and opening up the file navigation, the new grouping is there and works.
This is a CSRF issue.
015-05-14 09:40:15 WARNING plone.protect error parsing dom, failure to add csrf token to response for url http://localhost:8080/Plone/workspaces/open-market-committee/@@sidebar.default

gyst:
@pilz I've changed the storage. But there's no tests on the actual persistence....

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/15/
Test FAILed.

gyst:
Tests are green locally, jenkins is borked.

mattss:
@pilz master has been updated with some fixes for the robot tests. Please could you rebase this pr?

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/28/
Test FAILed.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/31/
Test PASSed.

pilz:
Thank you jenkins for the entertainment

pilz:
@gyst, how can you change a grouping on a folder? Once you navigate into a folder, changing the grouping is not possible anymore, the dropdown is hidden by the back button. How can I reproduce the protect issue? This suggests that somewhere there is a post involved, which shouldn't, I think

gyst:
@pilz Must've been an incomplete setup on my end, cannot reproduce now.
Looks fine to me. Have you reviewed my commit?
OK to merge with me.

PR #245, opened 2015-05-29
Attachments security

gyst:
Rip out the unprotected ++attachments++default traverser.
Instead, use security-protected views as documented in the docs.

@jaroel care to review?

@wengole this already weakens the docconv.view dependencies. You will probably want to review this and merge it into the async-final branch. The testcontent setuphandler still breaks on docconv async so I've kept that disabled.

This should be completely secure except for the documented theoretical situation, where:
- an attacker has an account in the portal
- but is not a member of a private workspace
- in which another user uploads an attachment

There's a small window where the attacker can theoreticaly access the attachment, if he can guess the username and timestamp and attachment filename and he does all of that before the temporary attachment gets moved to the definitive statusupdate annotation.
We will close that theoretical vector soon anyway when refactoring workspace View permissions.

I have a Heisenbug issue in the robot tests which should be unrelated but can give a red test result. There's a small change in this commit set for that which does not solve that completely.
Another small unrelated change is the personal toolbar image, result of an imperfect rebase resolution.


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/89/
Test PASSed.

PR #256, opened 2015-06-03
Zcml layer binding

gyst:
This atones for sins by omission mostly committed by @mattss but also @wengole @pbauer @pilz @adamcheasley and others.

Clean ZCML is not something optional. 
- Missing browser layer bindings cause weird bugs in multi-site setups, make it impossible to cleanly debug or re-use individual packages, and make it impossible to switch themes.
- Overly broad permissions and for="*" declarations can open up security holes.

Can somebody can please merge this, so I can get on with theme switching?

Also removes some stale theme views.


adamcheasley:
@gyst I'm not sure what you're stating by your first point. Are you saying that a layer directive should _always_ be added to a view/page/resource declaration?

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/102/
Test PASSed.

mattss:
This looks good and makes sense. We never run multi-site setups anymore due to performance, and we are usually customising existing code rather than writing code to be customised, which is why we often miss these. Layers will be crucial to customising and extending Plone Intranet for clients.

PR #261, opened 2015-06-04
Also inject the breadcrumbs when navigating the tree in the sidebar. …

### LINKED ISSUE #258: Workspace sidebar: entering a folder doesn't update the breadcrumbs
This is because we inject the @@sidebar.default tile when navigating to a folder. Ideally we would be able to also inject the breadcrumbs from another tile (but this isn't yet supported in patterns: Patternslib/Patterns#406) so we will need to use a view which includes both the sidebar tile and the breadcrumbs.

pilz:
…Fixes #258


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/106/
Test FAILed.

PR #263, opened 2015-06-04
Plone5b2 introduces a new bundle, which we need to exclude explicitly…

### LINKED ISSUE #247: beta2: file dropzone
Upgrading to beta2 gives a regression in the drag-and-drop upload area in the workspace navigation.

pilz:
… to avoid getting its css. Fixes #247
@cillianderoiste


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/108/
Test PASSed.

mattss:
I was wondering why this had happened. Is there any way to exclude all plone bundles, or will we hit this every time the default Plone bundles change?

pilz:
I haven't looked at the code, but I only know of the exclude option so far. Seems the default assumes you want to reuse barceloneta and therefore exclusion would be the exception

PR #264, opened 2015-06-04
Basic search

mattss:
This is an initial implementation of the search, including:
- Generic search API designed to allow multiple backend search implementations
- Initially backed by the portal_catalog
- Basic implementation of the 'all' search results tab (others to follow)
- Filtering by 'friendly type' (groups file content by their mime type descriptions)
- Filtering by tag
- Filtering by date range
- Robot-based acceptance tests
- This has been code-reviewed within Netsight, but please comment if you notice anything funny.

Work is ongoing to replace the backend with solr - which will form a future pull request.


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/110/
Test PASSed.

gyst:
I systematically get a test failure on this.

```
1/52 (1.9%) Manager can post a status update (posting.robot)                                        
```

Element locator 'css=textarea.pat-content-mirror' did not match any elements after 30 seconds

mattss:
I can't think of anything that this changes that would even relate to that test. Odd that jenkins doesn't get that.

gyst:
Some other remarks:
- test_friendly_name does not cover MimeTypeException code path
- test_zcatalog does not cover multi-tag searches, nor does search.robots
- test_zcatalog does not cover type or date searches but that's OK since the robot test covers that
- test_zcatalog does not cover empty searches, nor does search.robots

And:
- The documentation is very sparse. Relying on generated API docs (which is cool btw) is not enough. There needs to be a bit more about the reasoning behind the UX and the API i.e. core goals and core architecture choices. For example how the facets are drill-down on the primary search, how this is done with injection. And later, when we have more types of listing, how these are different from the "All" search and how an image or people search is different from an image or type drilldown on the All search.

gyst:
Test failure was probably a result of me having enabled CSRF protection locally in bin/test. Outside of the scope of this pull :-). Now testing master merge.

mattss:
@gyst thanks for the comments - I will try and improve the test coverage as suggested.

In terms of documentation, I can definitely elaborate on the API design decisions, but explanations of the prototype UI probably belong outside 'developer documentation'. Do we have any existing documentation about the prototype rationale for other areas of the system?

gyst:
@mattss None of those issues is a merge blocker but it would be great if some of that can be improved in the future.

Re docs: the developer documentation should at least not just tell me that there is an API, but why there is an API, how it's roughly structured and why it is structured that way. It should also tell me a bit about the custom indexers hook into the indexing process, and: why, goals! It should also tell me that the facets are drill-down searches that maintain the original search query and filter down on that. All of that is completely self-evident if you worked on the code for four weeks, but it'll be a very welcome high-level overview for somebody trying to understand the system from scratch. Can all be done in a few paragraphs, nothing very wordy.

Re docs I did a bit on UI mechanics in http://docs.ploneintranet.org/development/components/microblogstream.html#template-structure-and-injection. That's not crucial though, except for the fact that we're transmitting the original query as hidden along with any facet subrequest queries. That could take somebody a day to figure out if it's undocumented.

I usually write my documentation with my future self as intended audience. In 6 months or a year I'll have forgotten most implementation details, and a quick reminder of how it's all structured is a very useful reboot helper when you need to dive in again.

mattss:
@gyst thanks for the merge - I have added issues #271 and #272 assigned to me to record your feedback.

PR #287, opened 2015-06-11
Uninstall

pbauer:
Some noteworthy decisions I made:
- do not remove instances of workspace, workspacecontainer or todo since they might contain important data.
- only reset options (actions and registry) where the default is very likely to be the wanted state
- reset to barceloneta (we habe no way of knowing whether another theme was enabled before installing ploneintranet.theme)
- remove roles added by workspace and todo

On dependencies
- auto-uninstall all dependencies from the ploneintranet.\* namespace when unstalling ploneintranet.suite. 
- Other dependencies are only uninstalled when registered by hand (e.g. collective.externaleditor was just recently added as a the dependencies of ploneintranet.suite. I don't think it should be uninstalled when uninstalling the suite since it it's likely to be a part of a existing site).
- when uninstalling the suite we also uninstall the following dependencies of dependencies: slc.docconv, collective.workspace, collective.documentviewer. This is since it is very inlikely that they are required by another site
- when uninstalling other packages from the ploneintranet-namenspace (not the suite) their dependencies are _not_ automatically uninstalled since this is the expected behavior of packages.


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/165/
Test PASSed.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/169/
Test PASSed.

mattss:
@pbauer is it worth having somewhere in the test framework that uninstalls the suite and checks the result? Or at least just runs the uninstall to check we aren't trying to uninstall dependencies that have been removed.

mattss:
@pbauer also it looks like maurits has made a release with your changes to quickinstaller (3.0.9) - so it might be worth pinning to that and removing the manual checkout stuff: https://pypi.python.org/pypi/Products.CMFQuickInstallerTool/

pbauer:
@mattss the suite has a test for the full uninstaller (see https://github.com/ploneintranet/ploneintranet/blob/uninstall/src/ploneintranet/suite/tests/test_setup.py#L54).

mattss:
@pbauer ah super - sorry missed that

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/180/
Test PASSed.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/181/
Test PASSed.

PR #299, opened 2015-06-22
Solr search implementation

mattss:
Here it is! Solr search at last. To activate solr you need to manually build out with 'solr.cfg'.

There's a lot going on here and much of the complexity is in the test setup (surprise surprise). 
The default test setup, just run the zcatalog-backed tests, and will ignore all the solr tests. You can run the solr tests by building out with solr.cfg and running ./bin/test as normal (or ./bin/test -s ploneintranet.search.solr).

It is worth running 'make docs' and then reading through the 'Developer documentation > Components > Search' page before looking at any of this code, as it will help you understand the bigger picture.


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/208/
Test PASSed.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/209/
Test PASSed.

gyst:
@mattss Please review my extra commits.

Other than that I see the connection pool doesn't work, because the pool itself is being re-initialized (via adapter lookup) on each and every call in each and every request. I'd expect a significant speedup from caching either the connection itself or the connectionpool.

gyst:
The test coverage is quite limited. I would expect some integration/functional tests in the search/solr package that verify a full solr roundtrip at the python level. Additionally the robot tests in the suite should exercise solr and check solr-specific features, like the ngram substring matching I just activated and the spelling suggestions.

Spelling suggestions suggest the query I just made, that's weird and should be suppessed. Search for 'maxa' and you get the suggestion 'maxa'. Search for 'maxage' already gives you the 'manage' results btw.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/210/
Test PASSed.

mattss:
@gyst thanks for the feedback. We actually had started the ngram stuff in another branch I think, but I was just trying to get an initial PR in there before things got too complicated! 

In regards to connection pooling, I'm just looking through the code now and I'm not sure there even is any yet. There seems to be a _get_connection method but that is only used in the maintenance view. I will have to check this with @mgrbyte, as I'm not convinced we included connection pooling at this stage.

mattss:
@gyst regarding testing - all the solr interactions are handled by the scorched library (which cuts out a load of the manual data mangling that collective.solr does), and we are assuming that scorched's own tests would cover this kind of thing. Do you think we need to test that scorched is behaving as advertised, beyond what we have?

I agree that the robot tests are lacking with respect to solr, and this is related to an infrastructure discussion we may need to have separately regarding how we build the solr tests into jenkins, and whether or not we enforce a full solr setup on every PR. It might actually be simpler to put the solr robot tests into the ploneintranet.search package, rather than pollute ploneintranet.suite with all the solr setup/teardown gubbins.

mattss:
The spelling issues you are seeing are odd indeed. Ideally I would love to get this branch into master soonish so we can track/discuss these kind of issues out in the open, and avoid this branch getting stale.

mattss:
Remaining comments added as issues #302 #303 and #304

PR #308, opened 2015-06-24
User profile follow lists

mattss:
Implements follows/following portlets and tabs on user profile as per prototype


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/217/
Test PASSed.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/219/
Test FAILed.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/221/
Test PASSed.

gyst:
ploneintranet.api should not contain actual business logic, only nice delegation calls to the actual business logic. The stuff you're proposing here should be moved to userprofile and then be exposed in the API.

mattss:
@gyst I agree in principle, but the avatar_url method is really just a shortcut to plone.app.imaging. Or was it the existing 'create' method you are talking about?

gyst:
I think the get() and get_current() functions are clean, but the create() and avatar_url() functions contain all kinds of business logic that does not belong in an API. Stuff like image sizes and actually performing database actions (as opposed to delegating that) doesn't belong in an API IMO.

If you look at api.microblog.statusupdate (which should be just microblog...) I like the code there. It extends the existing backend API with extra datetime options but all of it is completely geared to API semantics, nothing else.

Simple fix would be to move create() and avatar_url() to ploneintranet.userprofile.somewhere and then import those back into ploneintranet.api.userprofile.

PS How do you feel about renaming ploneintranet.api.userprofile to ploneintranet.api.user?

mattss:
@gyst moving the methods elsewhere and then importing them into the api seems to just add a level of indirection that may confuse developers. We could move all the api methods back into an 'api' layer on each of the subpackages, but that removes the benefit of the single point of entry ploneintranet.api

This is turning into a discussion about what the API is actually for - because at the moment it is being used as a place to put methods that will be used by many other bits of code across the ploneintranet codebase - a combination of friendly entry-points to existing functionality, and 'utility' methods to avoid repeating common operations.

 We took plone.api as inspiration, but plone.api was initially created _after_ all the business logic that it uses in Plone which is why plone.api is so lightweight. In ploneintranet, the business logic does not exist yet as we are writing the api concurrently. However even for plone.api this situation is changing, as plone.api is growing with methods like 'transition' and 'find' having increasing amount of business logic in them. (See https://github.com/plone/plone.api/blob/master/src/plone/api/content.py)

gyst:
@mattss OK I get your point and we need to discuss that without blocking this PR.

I'm completely swamped and don't have time to actually test drive this, let alone resolve merge conflicts. If you can maybe ask @wengole to validate and merge this into master?

mattss:
@gyst thanks will do - I'll also create a ticket to follow up this api discussion.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/231/
Test PASSed.

PR #321, opened 2015-06-27
fix ?diazo.off=1

gyst:
There's more to this than just `<notheme>`. 
plone.app.theming also parses diazo.off and does magic with it.

### DIFF
--- a/src/ploneintranet/theme/static/rules.xml
+++ b/src/ploneintranet/theme/static/rules.xml
@@ -116,8 +116,6 @@
            css:if-content="#visual-portal-wrapper" />
 
 
-  <notheme css:if-content="body.diazo.off"/>
-
   <!-- These 3 rules are only needed untill these functionalities will be complete -->
   <drop css:theme="header #buddies-toggle" />
   <drop css:theme="header #notification-link" />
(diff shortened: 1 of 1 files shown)


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/249/
Test PASSed.

pilz:
I don't see either of these classes on the body, and adding diazo.off=1 reacts the same for me on master or this branch. How do I test this?

gyst:
I only verified that ?diazo.off=1 works, not what ?diazo-off does. Maybe it can be removed completely then.

pilz:
That diazo-off is, if anything, a class on the body. I added that in 9609c552214696315b4f8659165f721550d47fe0 but I just verified that in stock Plone5 b2 there is no such class. Perhaps it has been there once, otherwise I must have dreamt that up. Fact is that this rule doesn't get triggered anymore at all. So even if we remove it, diazo.off=1 works. I think we can remove it.

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/257/
Test PASSed.

gyst:
@pilz FYI just after merging myself this I discovered we actually need that rule. So I reverted the merge :-(

PR #325, opened 2015-06-27
fullname is calculated now, don't attempt to update it on subsequent run

pilz:
With the new user pofiles, the fullname is now calculated and can't be set anymore. Our setuphandlers code tries to set all profile attributes once again, if run a second time. By now, the new user profiles are applied and fullname can't be set anymore. Therefore we shouldn't attempt to.


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/255/
Test PASSed.

gyst:
OK I fixed that in library as well. Now I understand why :-)

PR #332, opened 2015-06-29
Events: use a custom widget for dates

cillianderoiste:
We need a custom z3c.form widget to save dates and times on events so that we can use the dexterity_update mechanism. Timezones are not yet supported.


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/287/
Test PASSed.

gyst:
Thanks @cillianderoiste looking good. Be aware you're globally registering adapters - it's safe in this implementation because they are only invoked by code that is bound to the IWorkspaceAppFormLayer.

PR #339, opened 2015-06-30
Restore action URLs in personal bar

mattss:
Looks like these were lost in a merge somewhere


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/293/
Test PASSed.

gyst:
Yes that was deliberate as part of the Mercury release, to not show links to unsupported pages.

mattss:
@gyst ah right - it also disabled the logout link! If we want to remove the 'social inbox' links then the corresponding portal actions should probably be removed/hidden instead.

PR #343, opened 2015-07-01
Tag auto suggest

cillianderoiste:
This is based on the personalized-tagging branch, so I guess it's best if @gyst can review it. Also, the robot test fails at the moment because the user doesn't have the correct permissions. I didn't look into this further but @gyst I guess you can take it from here?


gyst:
Jenkins test this please

gyst:
please test this test please this

mattss:
Test this please

mattss:
Jenkins get your act together and retest this please

mattss:
Test this please

cillianderoiste:
@gyst I was wrong on Friday, I thought that I got a JavaScript error when I returned unicode for both the 'id' and 'text' values, but it actually works fine. The "Member can tag a document with a tag suggestion" robot test now passes, but the others fail because of security errors so I'll pass it back over to you:

```
    ERROR:ploneintranet.network.browser.vocabulary:Vocabulary ploneintranet.network.vocabularies.Keywords lookup (subjects) not allowed
```

gyst:
Such failures are typically the result of trying to use the custom IDublinCore behavior in a setting where it has not been activated. i.e. this smells of an invalid test setup. If you changed the template for all of ploneintranet.workspace that test setup has to load ploneintranet.network:default. Where are those test failures documented?

gyst:
Jenkins please test this

mattss:
Retest this please Jenkins

cillianderoiste:
I'll look into hooking up ploneintranet.network for the tests.

cillianderoiste:
I noticed that suggestions for tagging images doesn't work in a full instance either, so it's not just the test setup. Images aren't using the custom IDublinCore behavior but I can register the custom CommaSeparatedWidget for the default IDublinCore behavior as well which causes them to be converted to tuples etc., but it doesn't help with the permission issue. I guess they should they also use the custom IDublinCore behavior, right?

To quickly reproduce the problem you could add a pdb to 
https://github.com/ploneintranet/ploneintranet/blob/tag-auto-suggest/src/ploneintranet/network/browser/vocabulary.py#L67
and see why `permission_checker.validate(field_name, factory_name)` fails for Images.

Any pointers on this @gyst, or could you take a look?

cillianderoiste:
Retest this please Jenkins

cillianderoiste:
Closing the PR so I can open one against master instead.

PR #348, opened 2015-07-01
Basic CSV upload for users

mattss:
This adds a browser view @@import-users to the profiles folder.
It attempts to validate the csv columns against all fields (and
behaviour fields) enabled on the user profile, and will give helpful
error messages if any are found.

Still to come - option to update existing users if matched by username.


pilz:
beautiful, thank you!

ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/322/
Test PASSed.

pilz:
I could test it with a csv generated from the given user excel sheet. 
I had to remove all colums which are unknown. I didn't find an address field, isn't that needed? The mechanism is very strict about the structure of the csv, which is good.

adamcheasley:
Currently the only columns that are required are those of the core fields, namely:
- username
- first name
- last name
- email

Everything else is optional, but each column in the csv _must_ map to a field on the userprofile which is either part of the core or has been enabled via a behaviour.

PR #349, opened 2015-07-02
connected "delete" buttons to delete_confirmation modal, refs #11910

reinhardt:
Known issue: The delete modal on the document view has a garbled header even though the markup of the link and the delete_confirmation page looks exactly the same to me as on the other views.

### DIFF
--- a/src/ploneintranet/workspace/basecontent/templates/document_view.pt
+++ b/src/ploneintranet/workspace/basecontent/templates/document_view.pt
@@ -32,7 +32,7 @@
                     <a href="${context/absolute_url}/external_edit"
                        tal:condition="context/externalEditorEnabled/available"
                        class="${view/icon_class} iconified">Open in External Editor</a>
-                    <a href="#confirm-delete" class="icon-trash iconified">Delete</a>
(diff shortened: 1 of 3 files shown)


ghost:
Can one of the admins verify this patch?

mattss:
Ok to test

mattss:
OK to test

mattss:
pat-modal is doing something odd that I don't understand, and adding the class of 'pat-modal' to every element on the injected delete confirmation markup, which is what screws the header up. @jcbrand any ideas?

jcbrand:
@mattss Yes, pat-modal adds the class "pat-modal" to content injected into the modal. I think the problem in your case is that multiple elements are being injected. So try to wrap them, or to change the pattern config such that a single element is inserted.

reinhardt:
What I don't understand is that it works for the event view but not for the document view. Both use the same link markup and are injecting the same content.

mattss:
Thanks for the update @jcbrand  -unfortunately as @reinhardt says, both event_view and document_view are using exactly the same markup, fetching exactly the same page for the modal, and yet get different results. I'm assuming there must be a conflict with something else on the document_view page.

reinhardt:
How about merging now and looking at the remaining issue later? I'd say it's much better to be able to delete stuff, even if the header looks weird.

mattss:
Agreed. Will merge for now.

mattss:
Add #363 to track header issues

reinhardt:
Cheers!

PR #354, opened 2015-07-02
add facility for switching between themes via UI action

gyst:
@pilz Here's your button.
Note there's a new registry record hostname_default that you'll have to configure and then re-run genericsetup for this to work properly.

No tests because it would be hellishly difficult to test host redirects on http://nohost.

### DIFF
--- a/src/ploneintranet/layout/browser/action.py
+++ b/src/ploneintranet/layout/browser/action.py
@@ -1,0 +1,27 @@
+from logging import getLogger
+from Products.Five import BrowserView
+from plone.app.theming.utils import theming_policy
+from ploneintranet.themeswitcher.interfaces import ISwitchableThemingPolicy
+
+log = getLogger(__name__)
+
+
+class SwitchThemeView(BrowserView):
+
+    def __call__(self):
+        """Redirect between default and fallback hosts.
+        Note that this is not protocol aware, so you'll have
+        to run either both on http or both on https."""
+        policy = theming_policy(self.request)
+        if ISwitchableThemingPolicy.providedBy(self.request):
+            log.error("@@switch_theme invoked but no switchable policy")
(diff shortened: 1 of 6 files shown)


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/330/
Test PASSed.

pilz:
I tried to apply only the actions.xml of layout and the registry.xml of themeswitcher to apply this to the current demo site, but importing the registry.xml garbles the theme.  I cannot see any reason why this should happen. Any idea? @gyst 
screen shot 2015-07-02 at 18 41 03

gyst:
On 02/07/15 18:54, Alexander Pilz wrote:

> I tried to apply only the actions.xml of layout and the registry.xml of
> themeswitcher to apply this to the current demo site, but importing the

The registry.xml of themeswitcher is completely barebones, intended for
generic re-use.

The actual registry.xml lives in ploneintranet.suite. Specifically that
one filters more browser layers.

BUT you will have a different registry.xml specifically for iKath. That
is the one you should run, since I've found out the hard way that
registry records should be considered to always run purge=True by
default. I.e. anything you run from stock ploneintranet will never be
valid for iKath that way.

So, todo:
- add hostname_default record to iKath registry.xml
- run iKath registry.xml

> registry.xml garbles the theme. I cannot see any reason why this should
> happen. Any idea? @gyst https://github.com/gyst
> screen shot 2015-07-02 at 18 41 03
> https://cloud.githubusercontent.com/assets/138906/8482700/a3b5c91c-20eb-11e5-9485-4d20e695f0c8.png
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/ploneintranet/ploneintranet/pull/354#issuecomment-118091748.

## 

```

```

PR #367, opened 2015-07-03
Update users via CSV

mattss:
This extends the existing bulk upload view with an 'update' flag, which
will update existing users if username matches (default is to raise a
duplicate username error).

Other changes:
- Transaction is aborted if any errors found, so the same CSV can be
  corrected and re-used
- Add count of number of users added
- Include helper list of available columns on the initial browser view


ghost:
http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/337/
Test PASSed.

pilz:
Looks very well and works fine. It just seems I can only make mandatory fields read_only. Optional ones don't react on the configuration.

pilz:
Works like a charm. Fired different broken csvs at it and it coped fabulously.

PR #373, opened 2015-07-06
show the images that are available as previews

pilz:
A small change to the library display. Don't use tag() which generates the height and width attributes. Css is setting these. Show a docconv preview, if one is available. And Show any lead image also in content, to avoid that users have to fiddle with the image once again.


pilz:
Checked with Cornelis, he is fine with showing the images on the detail pages. In the long run, he would like a checkbox on the news items/docs saying "Images visible as hero image on the news article page" and one saying "Image visible on news overview page". Default on for news items, default off for library pages.

PR #374, opened 2015-07-07
This does a few things for avatars

mattss:
1. Fixes a few places where avatars were missing (e.g. when mentioning
   users)
2. Reworks the avatar utility views to fit the multiple use cases in
   play


wengole:
Looks good

PR #391, opened 2015-07-08
Add support for the raptor image picker plugin

jcbrand:
This pull request mainly adds the RaptorImagePickerView (@@workspaceImagePicker) which is used by Raptor's image picker plugin to fetch and show images that may be inserted into the editor.

The image picker plugin is a "premium", non-free plugin, so we don't bundle it with PI, but this view at least adds support for it to integrators who would like to use it.

cc @pysailor


gyst:
No tests, no documentation.
@jcbrand can we please at least have a short pointer in the docs that this view is there, with instructions on how to buy and enable the plugin?

gyst:
@pilz says this is an incomplete proof-of-concept that needs debate.

gyst:
@pilz should I cancel this PR?

pilz:
No, I am about to acquire replacement js support that might pick that up soon.

pilz:
Cornelis' approach to a picker is clearly minimalistic while the pat-raptor approach is monolithic. Also it is commercial and the documentation is suboptimal. We will attempt to do a minimalistic picker ourselves and thus I close this.

PR #438, opened 2015-07-08
While a member creates a workspace, he is not yet given the TeamManag…

### LINKED ISSUE #388: As a user, I can open the add workspaces modal but not send it

pilz:
…er role, but he needs rights to access the workspace. One way to solve this is to give access contents information rights to the owner in the workflow. Fixes #388


gyst:
This was supposedly fixed before in
https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/workspace/subscribers.py#L57
It was specifically done that way so we can avoid assigning wide-ranging permissions to "Owner". Once cannot easily transfer "owner" unlike "team manager" and this is going to bite us in the future.

Isn't it possible to extract the group assignment from this "added" event handler 
https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/workspace/configure.zcml#L37
to an event handler earlier in the object creation cycle, like IObjectCreatedEvent http://docs.plone.org/develop/addons/components/events.html#creation-events

pilz:
Good question, I am not at all into these concepts. @mattss is that something that triggers a good idea in your head?

gyst:
Maybe this is because the security manipulation is done without elevating to superuser first!

gyst:
After stepping through Zope2 AccessControl for 2 hours I'm inclined to capitulate and accept this "Owner" workaround. The user is properly added to the 'Admins' group before the security exception is raised - I don't understand why the permissions is not properly obtained from the workflow via the role obtained from the group membership :-).

However, applying this patch, while fixing the Unauthorized, results in a different traceback:
2015-07-10 19:19:05 ERROR Zope.SiteErrorLog 1436555945.260.92648145826 http://localhost:8080/Plone/workspaces/add_content
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module ploneintranet.workspace.browser.add_content, line 27, in **call**
  Module ploneintranet.workspace.browser.add_content, line 90, in create
  Module ploneintranet.workspace.workspacefolder, line 57, in set_external_visibility
  Module plone.api.content, line 2, in transition
  Module plone.api.validation, line 70, in wrapped
  Module plone.api.content, line 2, in transition
  Module plone.api.validation, line 131, in wrapped
  Module plone.api.content, line 2, in transition
  Module plone.api.validation, line 100, in wrapped
  Module plone.api.content, line 416, in transition
  Module Products.CMFCore.WorkflowTool, line 234, in doActionFor
WorkflowException: No workflow provides the '${action_id}' action.

This is not seen by Jenkins because this PR does not add test coverage on the actual bug...

gyst:
Since both Alex and I got stumped, reassigning tot Matt.

mattss:
I'm on it

mattss:
Brief update here - adding a workspace as a non-manager works just fine if you manually go to plone/workspaces/++add++ploneintranet.workspace.workspacefolder. The workspace is created, and you are given the TeamManager role as expected.

The issue here appears to be specific to the way we are creating workspaces in PI - via our custom 'addcontent' view, which uses plone.api.content.create under the hood. As far as I can tell, the error occurs during a 'rename' which happens during the creation. Investigation continues.

mattss:
@gyst @pilz I've now found a resolution for this. It's not too pretty, but other than completely refactoring the whole workspace creation logic, this seems like the best way forward. Basically we have to re-initialise the zope security manager part way through the request, so that we have the permission to rename/transition the workspace that we've just created.

PR #459, opened 2015-07-08
updated diazo to get the recent branding in. This fixes also some UI …

pilz:
…issues like the Hamburger nav


gyst:
The portal tabs should be collapsed when you're in a workspace until you click the hamburger. In this implementation they're always expanded.

pilz:
Are you sure? This uses pat-store, so your choice is remembered once you take it. 
If I use a completely reset browser (no website data anymore) and even log in as a user I never used before, I see the portal tabs, until I step into a workspace. Then they are collapsed. I can open them again and if I step out and into the workspace again, they will still be open - until I manually close them again - or clear my browser.

gyst:
I was not aware of pat-store. Re-tested and you're right of course.

PR #463, opened 2015-07-09
Document @@switch_theme and move implementation into themeswitcher

gyst:
While documenting I found that @@switch_theme belongs in themeswitcher, not in layout.


pilz:
works fine

PR #466, opened 2015-07-09
override ../view url for library files in search results

gyst:
Link to direct download rather than file/view in search results.

### DIFF
--- a/src/ploneintranet/library/browser/baseviews.py
+++ b/src/ploneintranet/library/browser/baseviews.py
@@ -1,5 +1,6 @@
 from logging import getLogger
 from plone.memoize import view
+from Products.Five import BrowserView
 
 from ploneintranet.library.browser import utils
 from ploneintranet.library.browser.views import LibraryBaseView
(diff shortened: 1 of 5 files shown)


gyst:
Jenkins retest this please

pilz:
The code looks valid, but if I create a new portal and upload a docx to the library, the search links to /view and I get an empty page (see screenshot). Manually adding /documentviewer indeed downloads the file, but apparently a default install doesn't forward /view to /documentviewer anymore :(
screen shot 2015-07-13 at 21 15 27

pilz:
You thought you get away so easily? ;) Now when I add a file object, select a file from my harddisk and save, I don't get the file view in barceloneta. Instead it immediately downloads the file again… 

I guess we need a way to apply that layer only to the PI theme, but not barceloneta.

gyst:
@pilz ready for merge

mattss:
Can I just clarify that linking directly to the download is always the desired outcome from search results?

I seem to remember something in Plone (Archetypes?) that linked to the 'view' of a file if you had permissions to edit it, but just downloaded it if you didn't.

Otherwise users will not be able to change details of a file item (title/description) that they find via the search, which could be frustrating.

One option that we discussed a while back was having an separate 'direct download' link in the search results for items that support it, although that may need to be saved for future development.

gyst:
@mattss The intent is to link to the proper view of the file as defined in it's containing IApp. For files in workspaces that is the defined view there. For files in library that is the direct download view. For files in Barceloneta, wherever they live, it's the default plone behaviour described by you.

So by binding to IAppContent specific library layers here, I'm implementing direct download only for files contained within an ILibraryApp which are viewed through ploneintranet.theme. All the other cases are left alone.

mattss:
@gyst Thanks for the clarification. Makes sense now.

PR #467, opened 2015-07-09
Reenable news portlet on dashboard

pilz:
Now that we have news items again in the library, the news portlet makes sense again and can be shown.


pilz:
jenkins please retest this

pilz:
retest this please

pilz:
test this please

pilz:
ok to test

pilz:
jenkins retest this please

pilz:
retest this please jenkins

src/ploneintranet/workspace/browser/tiles/workspaces.py
```diff
@@ -41,7 +41,7 @@ def my_workspaces(context):
         path=ws_path,
     )
     workspaces = []
-    for brain in brains:
+    for brain in brains[:5]:
         css_class = escape_id_to_class(brain.getId)
```
gyst, line 45: This change not only causes a test regression but is actually invalid. It makes it impossible to reach my workspaces if I'm in more than five.

pilz, line 45: Isn't that box only meant to give a quicklink style entry to the recent ones? If I don't find it there I can still go to workspaces and find it there. A filter by "my workspaces" may be an option there.

And if not, what about our managerial colleagues who can see all workspaces? They will have a list with 100 items in there? (Asking because I've been there)

gyst, line 45: Isn't that same function used in both the sidebar listing and the full workspaces listing?

gyst, line 45: If you move the [:5] slice up to line 25 the main workspace listing will not be affected.

pilz:
I actually prefer removing the limitation to 5 as it is not yet clear if that was really by intention in the prototype.

PR #471, opened 2015-07-10
Dexterity update handle no value

cillianderoiste:
This is a follow up from #390 which was merged and then reverted. It needs further work and testing before it can be merged. The problem it addresses is that submitting the correct empty value for a field (NO_VALUE) is currently ignored. This means that some fields can't be reset. We have this problem with a field we added to iKath (registration_number).

The current problem with this PR is that a POST request to the (edit) view of an item will cause any fields not included in the request to be set to the default. Perhaps it is enough to only update the fields which are in the request.form (in dexterity_update).

### DIFF
--- a/src/ploneintranet/workspace/basecontent/baseviews.py
+++ b/src/ploneintranet/workspace/basecontent/baseviews.py
@@ -63,7 +63,7 @@ class ContentView(BrowserView):
             modified = True
             messages.append("The workflow state has been changed.")
 
-        if self.can_edit:
+        elif self.can_edit:
             mod = False
             if self.validate():
                 mod, errors = dexterity_update(context)
--- a/src/ploneintranet/workspace/basecontent/templates/document_view.pt
+++ b/src/ploneintranet/workspace/basecontent/templates/document_view.pt
@@ -87,10 +87,14 @@
                     <img src="${python:context.absolute_url()}/docconv_image_preview.jpg?page=${preview}&salt=${preview_hash}" />
                   </tal:previews>
                 </article>
+                <tal:file_nochange condition="python: context.portal_type == 'File'">
+                  <input type="hidden" name="file.action" value="nochange"/>
+                </tal:file_nochange>
 
                 <article class="document preview" tal:condition="python: context.portal_type == 'Image'">
                   <figure>
                     <img src="${view/image_url}" title="${context/title}" alt="${context/title}" />
+                    <input type="hidden" name="image.action" value="nochange"/>
                   </figure>
                 </article>
               </div><!-- #document-content -->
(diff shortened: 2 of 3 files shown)


gyst:
Jenkins retest this please

gyst:
Please provide integration test coverage, this is key functionality that is exercised a lot.

reinhardt:
If it was only a matter of updating the tests or adding new cases I would agree that this has to be done first. However it looks like there never have been any tests for dexterity_update, so I would argue that it is a separate task to write the missing tests. I would merge this now and then add a ticket for writing tests for dexterity_update. @gyst, does this make sense?

gyst:
Yes Manuel ok.

On August 10, 2015 2:34:28 PM GMT+02:00, Manuel Reinhardt notifications@github.com wrote:

> If it was only a matter of updating the tests or adding new cases I
> would agree that this has to be done first. However it looks like there
> never have been any tests for dexterity_update, so I would argue that
> it is a separate task to write the missing tests. I would merge this
> now and then add a ticket for writing tests for dexterity_update.
> @gyst, does this make sense?
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/ploneintranet/ploneintranet/pull/471#issuecomment-129427750

## 

Sent from my Android device with K-9 Mail. Please excuse my brevity.

gyst:
Can we please get a bug reproduction scenario for this in vanilla plone intranet? As it is, it's impossible for me to either merge or work on this. @reinhardt what is your manual test scenario?

reinhardt:
I hope @cillianderoiste can expand on this, but the test case is basically setting a value for a field, then removing the value. The new empty value is not stored but the old value retained. AFAIK only certain field types are affected.

gyst:
On 28/08/15 09:44, Manuel Reinhardt wrote:

> I hope @cillianderoiste https://github.com/cillianderoiste can expand
> on this, but the test case is basically setting a value for a field,
> then removing the value. The new empty value is not stored but the old
> value retained. AFAIK only certain field types are affected.

OK. And what is the reason @pilz reverted the earlier PR containing the
same changeset as this one? We need a scenario for that reason as well.

## :*CU#

```

```

cillianderoiste:
In iKath we have fields which use vocabularies which cannot be unset. I can cook up a test for this.

In the version that @pilz reverted, posting an empty request would reset all fields. In the current version only fields which are in the request are modified.

cillianderoiste:
Sorry for the noise. I think there is no need for this. We were using a select with a Choice field which was missing value="--NOVALUE--" for resetting the field, and we were also using a Choice field for a text input with pat-autosuggest. I'll close this PR.

PR #477, opened 2015-07-10
Correctly check for missing catalog values

mattss:
This code was meant to exclude blank/Missing.values from the lists of facets generated in the zcatalog search, but it was checking the brain itself rather than the column/metadata value!

FYI @cillianderoiste


gyst:
Jenkins retest this please

mattss:
Woohoo jenkins comments seem to be working again!

Issue #371, opened 2015-07-06
notifications MVP: mentions


gyst:
As a user,
When another user @mentions me in a post,
I'd like to get notified of that,
So that I can respond and sustain meaningful conversations with my colleagues.

As a user,
When I mention another user in a post,
I expect them to be notified of that,
So that I can use mentions instead of having to send an email.

The notifications subsystem is potentially a big system. A full fledged event listening infrastructure needed to realize e.g. #360 (show "published" events) raises all kinds of architectural questions (like async for "exploding" events and how to summarize truckloads of similar notifications). @mattss rightly suggested we need a sprint to formulate and execute a coherent plan for that.

This ticket is of a much smaller scope. Instead of anticipating the full event listening system described above, this is much simpler: if a user posts a statusupdate that mentions one or more others, use the notification box to show those others that they were mentioned.

A minimal implementation could work as follows:
- mentions are already stored on StatusUpdate objects on creation (verify)
- on handling a statusupdate post (in the statuscontainer.add() method), for each mention, create a corresponding notification for the mentioned user, and store this notification
- on loading a page, the notification box shows the unread notifications count
- on opening the notification box, it shows a list of notifications and automatically marks all notifications as read

NB:
- there's no need for an "all notifications" view at this stage
- a notification of a mention links directly to the relevant post
- there's no need for websocket push to update notifications and unread counter, a page refresh is fine

The upside of this is that this builds the initial notifications roundtrip including backend/frontent integration. We can then later embellish that with async handlers, websockets etc etc.

mattss:
Work in progress on https://github.com/ploneintranet/ploneintranet/tree/basic-notifications

mattss:
@gyst Is there a way to view individual status updates? How do I link from the notifications listing to the status update the notification is for?

gyst:
On 10/07/15 13:47, Matthew Sital-Singh wrote:

> @gyst https://github.com/gyst Is there a way to view individual status
> updates? How do I link from the notifications listing to the status
> update the notification is for?

There is a StatusUpdateView in activitystream but it's a helper view
that IIRC is not directly accessible on a URL yet.

https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/activitystream/browser/statusupdate.py

It should not be that hard to subclass that into an actual traversable
view on the navigation root that renders an individual statusupdate, or
rather the whole conversation that statusupdate is a part of.

[...]

PR #483, opened 2015-07-10
Basic notifications for status updates

mattss:
See https://github.com/ploneintranet/ploneintranet/issues/371 for spec.

I've tried to re-use as much of the existing ploneintranet.notifications framework as possible in the interests of time, but I am not convinced the current architecture is quite right for the future.

Still to do - some way to link directly to status updates from the
notifications page


gyst:
Opening the notifications box marks notifications as read, which is cool. What is awkward though, is that the unread counter doesn't change. So you keep having this red notifier of "unread" messages and when you click it, it says "no unread notifications".

Even in the absence of websockets push, this can be fixed by, on opening the notifications box, not only injecting the notifications list but also resetting the unread marker as a secondary injection.

gyst:
Helper view follow-up issue #486

mattss:
@gyst At the moment the notifications popup is using an ajax pat-tooltip, rather than pat-inject directly, so I'm not quite sure how or if you can add secondary injection options. I could experiment with changing it to a pat-inject, but I'm a bit concerned about straying too far from the prototype.

gyst:
@mattss Various issues:
- I've tried to fix some URLs in the notification but am stumped by the actual message url. Leaving that empty results in the message pointing to the current page (an A tag without HREF) which is invalid if you're accessing the notification box from a page other than the dashboard. You can see that in action if you go to /notifications?showall=1.
-  There's another bug there in showall: the menu is not diazo injected like it should and displays the proto menu item. E.g. dashboard URL should be "/" not "/dashboard.html".
- Finally: security. If you activate a view that has a security comment you should check against anon access. In this case that results in a traceback. I'd rather have that handled explicitly with a check against anonymous and a redirect to "/" if so.

mattss:
@gyst I think i've resolved the issues you were having:
- URLs - updated the template to use the url that you added (to the root of the site)
- Menu - the notifications template was quite old, so I've updated the relevant classes and tweaked the diazo rules to fix this
- Permissions - tied them to ploneintranet.viewsite as per other site-level views

gyst:
Looking good, thanks!

PR #495, opened 2015-07-13
Fix pep8 problems

ale-rt:
This prevents developers to make any further commit

```
ale@kenobi ploneintranet]$ ./bin/flake8 src/ploneintranet/
src/ploneintranet/workspace/basecontent/baseviews.py:124:16: E713 test for membership should be 'not in'
src/ploneintranet/workspace/basecontent/widgets.py:2:1: F401 'IDublinCore' imported but unused
```


ale-rt:
An identical fix was applied in eb965d0a2ee7668ea1181b6b9db256803f76957d

Issue #21, opened 2015-03-13
AttributeError: Options instance has no attribute 'query_bool'


wengole:
<a href="https://github.com/macagua"><img src="https://avatars.githubusercontent.com/u/185395?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [macagua](https://github.com/macagua)**
_Friday Nov 21, 2014 at 22:47 GMT_
_Originally opened as https://github.com/ploneintranet/ploneintranet.suite/issues/13_

---

Hi guys 

When I try to build a instance, and show me the following error:

```
Installing 'collective.recipe.omelette'.
We have the distribution that satisfies 'collective.recipe.omelette==0.15'.
Version conflict while processing requirement zc.buildout (constrained to zc.buildout==1.6.3)
Getting required 'zc.buildout==1.6.3'
We have the distribution that satisfies 'zc.buildout==1.6.3'.
Version conflict while processing requirement setuptools (constrained to setuptools==0.6c11)
Getting required 'setuptools==0.6c11'
We have the distribution that satisfies 'setuptools==0.6c11'.
Installing 'z3c.recipe.scripts'.
Picked: z3c.recipe.scripts = 1.0.1
Version conflict while processing requirement setuptools (constrained to setuptools==0.6c11)
Getting required 'setuptools==0.6c11'
We have the distribution that satisfies 'setuptools==0.6c11'.
Version conflict while processing requirement zc.buildout>=1.5.0 (constrained to zc.buildout==1.6.3)
Getting required 'zc.buildout==1.6.3'
We have the distribution that satisfies 'zc.buildout==1.6.3'.
While:
  Installing.
  Getting section scripts.
  Initializing section scripts.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/home/macagua/.buildout/eggs/zc.buildout-2.2.5-py2.7.egg/zc/buildout/buildout.py", line 1946, in main
    getattr(buildout, command)(args)
  File "/home/macagua/.buildout/eggs/zc.buildout-2.2.5-py2.7.egg/zc/buildout/buildout.py", line 510, in install
    [self[part]['recipe'] for part in install_parts]
  File "/home/macagua/.buildout/eggs/zc.buildout-2.2.5-py2.7.egg/zc/buildout/buildout.py", line 1098, in __getitem__
    options._initialize()
[...]

PR #496, opened 2015-07-13
Improve bootstrap experience

ale-rt:
Having an up to date bootstrap.py helps avoiding buildout issues (e.g. #21).

I have updated bootstrap.py to the latest version which supports this syntax:

```
python bootstrap.py --buildout-version=2.2.5 --setuptools-version=14.0
```

I also update other related files.

I noticed that there is a requirements.txt file, but it's usage is undocumented.

I am also a fan of pip installing zc.buildout, but I think we should ship just one documented way.

What do you think about it?


gyst:
@ale-rt The currently used practice (see Makefile: make devel) is to ignore bootstrap.py. Instead:
- virtualenv .
- bin/pip install -r requirements.txt
- bin/buildout

I'm surprised our docs are incorrect for that.

ale-rt:
I am really glad requirements.txt is preferred!

I will update the documentation.

To avoid confusion:
- I would like to remove bootstrap.py
- modify .travis.yml to be consistent with this practice

I would also upgrade:
- zc.buildout to 2.4.0
- setuptools to 18.0.1

Please @gyst, stop me if you do not agree on some of this points.
Thanks!

PR #502, opened 2015-07-13
Members can't change sidebar grouping

pilz:
This fixes an issue where non admins couldn't change the sidebar grouping.


pilz:
Jenkins green again.

PR #505, opened 2015-07-13
add detailed checks for each content type's personal permission

pilz:


src/ploneintranet/workspace/browser/tiles/sidebar.py
```diff
@@ -70,12 +70,80 @@ def can_manage_workspace(self):
...
+        Cave. This was an easy way when we had archetypes.
+        With Dexterity, each content type is protected by its own
+        permission. Use more specific checks, if you can.
         """
         return api.user.has_permission(
             "Add portal content",
             obj=self.context,
         )
 
+    def can_add_documents(self):
```
ale-rt, line 83: Can we add a @memoize decorator from plone.memoize.view?
Those methods are called many times in the template.

pilz:
Of course!

ale-rt:
Awesome work!

Issue #449, opened 2015-07-08
workspace Participant policy Consume - user has still all add buttons and they fail


maartenkling:

allcaps:
The button should stay. It's a bug deep in security. Ask @gyst.
The real bug: A normal user should be able to add a workspace.

gyst:
Sorry for the confusion. Adding a workspace is #438 - this plays out in the "workspaces" container view.

This ticket is something different: in a Consume workspace a user has no right to add content, hence should have no "add document", "add folder" buttons. Possibly already fixed in #505 please check.

gyst:
Fixed in #508

Issue #407, opened 2015-07-08
workspace: security change is slow and notification is done wrong


maartenkling:

maartenkling:
i need to refresh to get the notification

pilz:
I heard during dinner in AMS that we don't use experimental.securityindexing yet. Might that be related?

pilz:
The need to refresh points to an injection issue.

gyst:
Fix in #508

PR #508, opened 2015-07-14
449 workspace ui permissions

gyst:
Tries to fix workspace UI permissions, undoing the fine grained permission checks and using only the workflow managed cmf.AddPortalContent.

Doesn't work though. May indicate a deeper security issue in the "Consume" policy.

### DIFF
--- a/src/ploneintranet/workspace/browser/tiles/sidebar.py
+++ b/src/ploneintranet/workspace/browser/tiles/sidebar.py
@@ -71,12 +71,23 @@ class BaseTile(BrowserView):
     def can_add(self):
         """
         Is this user allowed to add content?
-        Cave. This was an easy way when we had archetypes.
-        With Dexterity, each content type is protected by its own
-        permission. Use more specific checks, if you can.
+        Cave. We don't use the plone.app.contenttypes per-type permissions.
+        Our workflows don't map those, only cmf.AddPortalContent.
         """
         return api.user.has_permission(
             "Add portal content",
+            obj=self.context,
+        )
+
+    @memoize
+    def can_edit(self):
+        """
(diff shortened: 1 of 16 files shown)


gyst:
Fixes #449. Also fixes #407 - the event handler there did never work and the injection was borked as well. Follow-up ticket #510 should fix that member groups changes are not persisted across instance restarts!

src/ploneintranet/workspace/browser/tiles/sidebar.py
```diff
@@ -250,6 +190,7 @@ def update_field(field_name):
                 )
 
         if self.request.method == 'POST':
+            self.form_submitted = True
```
gyst, line 193: This fixes the injection

src/ploneintranet/workspace/configure.zcml
```diff
@@ -47,8 +47,7 @@
   <!-- Subscribe to ParticipationPolicyChangedEvent in order to
        update existing users-->
   <subscriber
-     for=".workspacefolder.IWorkspaceFolder
-          .interfaces.IParticipationPolicyChangedEvent"
+     for=".interfaces.IParticipationPolicyChangedEvent"
```
gyst: This fix is needed to get the subscriber actually called!

mattss:
@gyst - added my fixes for #510 in here. Also reverted your subscriber change. Can't explain why you had to do that I'm afraid, but reverting it makes all tests pass, and UI sidebar still fires the event in my manual testing.

mattss:
I've reviewed the changes made by @gyst in this and I'm happy that the buttons now appear to match the assigned policy.

Someone may want to review my changes before merging.

Issue #506, opened 2015-07-14
StatusUpdate: rename the context property to microblog_context


ale-rt:
Duplicates part of #359

ale-rt:
This can be closed by merging #511

gyst:
Merged.

PR #509, opened 2015-07-14
Use piapi to work on the microblog context

ale-rt:
I did a little bit of refactoring to the microblog context.
This is a first step for #506


ale-rt:
@gyst can we merge this first step?

PR #511, opened 2015-07-14
506 context is microblog context

ale-rt:
This contains:
- the refactoring of #509
- rename of the property context of the StatusUpdate
- an upgrade step

We should notify everybody to launche the upgrade step for pi.microblog in the plone control panel.


src/ploneintranet/microblog/statusupdate.py
```diff
@@ -32,8 +27,14 @@ class StatusUpdate(Persistent):
         IStatusUpdate,
     )
 
-    def __init__(self, text, context=None, thread_id=None, mention_ids=None,
-                 tags=None):
+    def __init__(
+        self,
+        text,
+        context=None,
```
gyst, line 33: This should already become microblog_context in the constructor

ale-rt, line 33: Agree, I will update the PR if it is not merged as is.

gyst:
We need a migration step for this to protect our launching customer's database integrity.

Issue #410, opened 2015-07-08
http://localhost:8080/Plone/@@personal-preferences unstyled


maartenkling:
as a user i can reach this page so it needs to be styled or removed

gyst:
Link should be hidden or link to editable userprofile

allcaps:
The view should be blocked. @@personal-preferences should give 404.

gyst:
Better is to redirect to your actual user profile, like @@author does.

gyst:
The view now returns a 404 as per #513. 
Todo:
- add 'Personal profile' action pointing to actual user profile as per prototype
- do not add 'Personal settings' since we don't have that implemented yet
- add logout actions

The markup should implement exactly the prototype so it looks like the prototype, see below.
This may require some diazo tweaking in ploneintranet.theme after setting up the portal actions in ploneintranet.suite

![actions](https://cloud.githubusercontent.com/assets/122351/8683449/398d7a3a-2a73-11e5-9b89-1997d89d06c2.png)

ale-rt:
In PR #530 I implemented the menu from prototype:
1. My Profile points to http://localhost:8080/Plone/@@my-profile
2. Personal settings points to http://localhost:8080/Plone/@@my-profile#person-info (and works fine opening the right tab)
3. Logout points to http://localhost:8080/Plone/logout

The three functionalities are covered by robot tests.

IMHO we should merge #530 (once jenkins is fine with that) and close this issue, given that @@personal-preferences has been turned of by #513

PR #513, opened 2015-07-14
Switchoff some pages

mauritsvanrees:
Up for discussion.
Some more pages may need to be added.
But let's first see if this is the right direction.
Not sure if there is a ticket related to this.


gyst:
Fixes #410 partially - change of portal action there still todo.
Fixes a lot of non-ticketized issues.

Issue #338, opened 2015-06-30
Implement UI to assign Roles to Members and remove them


pilz:
We can now add members to workspaces but not remove them or assign roles.

reinhardt:
Implementation done.
To do: robot tests.

reinhardt:
Robot tests added, see pull request #518

PR #518, opened 2015-07-14
338 manage ws members

reinhardt:
UI for managing workspace members and their roles.
- add multiple members
- remove members
- assign/change roles
- remove roles
- display roles
- hide functions for unprivileged users

Compare with prototype.


gyst:
Lots of security-sensitive feature work, but no test coverage at all. Please add some robot test coverage that:
- validates the extra roles visually
- but also validates the extra permissions actually, for example setting a workspace to 'consume', then elevating a single user, then validate that this user has gained 'add' permission

See for example the work I did in
https://github.com/ploneintranet/ploneintranet/commit/347a99d7b4bc750ce38f9de866377d12bf98e81f

reinhardt:
I'm not sure the second part is within the scope of this. I just connected UI elements to existing role management features. The latter may or may not be tested - I didn't touch them.

That said, robot tests that verify the UI are still to do, as mentioned in #338.

gyst:
@reinhardt You don't have to do heavy checks. Just something that not tautologically tests your own changes only. If you're changing permissions, other elements in the UI should change. Anchoring the test on one of those changes is sufficient.

gyst:
@reinhardt please also merge in master so I can auto-merge on github once this is complete.

reinhardt:
Added two robot tests and rebased/merged master.

reinhardt:
...and two more.

src/ploneintranet/workspace/browser/tiles/sidebar.py
```diff
@@ -127,22 +136,60 @@ def users(self):
...
+                        _(u'Member(s) removed'),
+                        self.request,
+                        'success',
+                    )
+                elif batch_function == 'role':
+                    role = self.request.get('role')
+                    default_role = ws.participant_policy.title()
+                    groups = [default_role]
+                    if role:
+                        groups.append(role)
```
gyst: This is in line with our original technical design, but it conflicts with the UI design. This adds custom roles on top of the default workspace roles. The UI clearly indicates that you're reassigning the user from the default role to another role.

In this implementation, if you're in a 'publish' workspace you can move users to 'consume' visually but it has no security effect whatsoever, because the user is in both ('publish', 'consume') groups and publish is a superset of consume.

The reason that this problem doesn't show up in the tests, is because they are run in a consume workspace and elevate permissions.

To summarize: with this implementation you can only elevate permissions, not assign lesser permissions than the default. IMO that should be changed so that a custom role _replaces_ rather than _adds to_ the default group.

@mattss do you agree?

src/ploneintranet/suite/tests/lib/keywords.robot
```diff
@@ -392,6 +439,11 @@ I cannot see the document
     Click link  Documents
     Page should not contain  ${title}
 
+Allan has the option to create a document
+    I am logged in as the user allan_neece
+    I go to the Shareholder Information Workspace
```
gyst, line 468: This is a 'consume' workspace which is the lowest privilege level. That's OK for testing privilege enhancement, but not for reducing access rights. For that you'd need a different workspace with a higher default policy. See other comment.

gyst:
@reinhardt This is great work, thanks for the care you put in finishing and testing. I'm sorry it is your bad luck to expose a design inconsistency that I should have caught earlier.

src/ploneintranet/workspace/browser/tiles/sidebar.py
reinhardt: > The UI clearly indicates that you're reassigning the user from the default role to another role.

Interesting - to me it doesn't. The fact that some users have this orange bubble with a role while regular users don't have a bubble to me conveys the idea that these users have something extra on top of what the other users have. If all other users had a bubble with the default role, maybe in a different colour, that would in my eyes be closer to what you describe.

reinhardt:
Don't worry about it, always happy to stir things up! ;-) And thanks for the appreciation.

src/ploneintranet/workspace/browser/tiles/sidebar.py
gyst: I get your point, visually. However the wording of the interactions makes it clear this is about assigning roles, not assigning _extra_ roles. Probably fixable by replacing the default group with the new one, instead of setting both.

reinhardt: > However the wording of the interactions makes it clear this is about assigning roles, not assigning extra roles. 

True.

> Probably fixable by replacing the default group with the new one, instead of setting both.

Will do (assuming @matss doesn't veto).

mattss: I _think_ I just broke the ability for policy exceptions in a recent commit (https://github.com/ploneintranet/ploneintranet/commit/06ac676d57546a66e99d389475f3ba2da61dc6c8) (sorry!)

Previously, this only moved uses from the currently policy to the new policy, but I changed it to move _all_ members to the new policy, because we were having persistence issues. I now think this move was foolish, so we may need to switch back to the logic where it only moves users who were in the old policy group (see diff for what it was doing originally).

If you want me to try and fix that, please let me know.

gyst: I think the new code is much more robust, and, as I note, it's also in line with how Cornelis has structured the UI. Simplicity is our mantra. Please keep the current behavior and let's do "replace" instead of "add" group roles.

gyst: Oops hold. I now see our testcontent does set multiple roles. Lemme think.

gyst:
So:
- make roles replacing not additive
- fix the collective.workspace permission problem
- also show 'Team Manager' as a role

gyst:
@reinhardt Matt will take over now.

reinhardt:
Roger that.

gyst:
Great work all!

gyst:
RRgh the merge broke the test. Will fix that. Later.

pilz:
I am checking the tests now

gyst:
@pilz I fixed this!

pilz:
lol :) Yes, I found no errors in the tests ;) Super!

PR #533, opened 2015-07-15
Add name attributes to checkboxes to pacify parsley.js

jcbrand:
Parsley.js throws an error when it parses checkboxes without name attributes, which stops our JS from evaluating further.

I looked up what appeared to be the correct names (i.e. they should correspond to the schema fields on the event).


jcbrand:
The failure is a WebDriverException, doesn't appear to me to be related to the change made.

http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/463/console

Perhaps related to  #534?

gyst:
Jenkins re test this please.

Issue #420, opened 2015-07-08
as a user when visiting a profile my breadcrums give me Profile link when hitting this get unauthorized


maartenkling:
http://localhost:8080/Plone/profiles/christian_stoney
Hit Profiles
http://localhost:8080/Plone/acl_users/credentials_cookie_auth/require_login?came_from=http%3A//localhost%3A8080/Plone/profiles

gyst:
This is because the profiles folder is private but should be published.
Partially duplicates #413 and #446

PR #541, opened 2015-07-15
Ensure we have a folder called profiles and it is published

### LINKED ISSUE #393: Cannot create a user to login
there is no way to create a user for the intranet

ale-rt:
Fixes: #420
I am not what to do with #393 after this is merged :)


gyst:
See my comment on #393 - todo there is only documentation.

PR #551, opened 2015-07-16
Check Delete objects permission before showing the icon bin

ale-rt:


ale-rt:
@gyst do you think checking "Delete objects" is enough?

gyst:
@ale-rt Yes, I've updated the docstrings.

PR #575, opened 2015-07-17
567 robot search timings

### LINKED ISSUE #567: Test for search and filter by content type fails randomly
The search test case "Allan can search and filter by content type" sometimes fails randomly, As an example: 

http://jenkins.ploneintranet.net/job/Plone.Intranet.Pull.Request.Builder/486/consoleFull#1659009341d7588e9e-566a-4586-b220-5d422b09bef9

but this has happened a few more times this sprint week on other pull requests that are completely unrelated to search.  Probably a timing that has to be fixed somewhere.

fredvd:
Fixes #567.

### DIFF
--- a/src/ploneintranet/suite/tests/acceptance/search.robot
+++ b/src/ploneintranet/suite/tests/acceptance/search.robot
@@ -31,9 +31,8 @@ Allan can search and find multiple documents
 Allan can search and filter by content type
     Given I am logged in as the user allan_neece
     I can search in the site header for minutes
-    Unselect Checkbox  css=input[type="checkbox"][value="Page"]
-    Wait Until Keyword Succeeds  1  3  Page should not contain  /workspaces/open-market-committee/manage-information/minutes-overview
-    Wait until page contains  /workspaces/open-market-committee/manage-information/minutes/view
+    Then I can exclude content of type Page
+    And the search results do not contain /manage-information/minutes-overview
 
 Allan can search and filter by date
     Given I am logged in as the user allan_neece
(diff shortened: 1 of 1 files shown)


ghost:
Can one of the admins verify this patch?

mattss:
Test this please jenkins

src/ploneintranet/suite/tests/acceptance/search.robot
```diff
@@ -80,11 +80,6 @@ I can follow the search result ${SEARCH_RESULT_TITLE}
     Click Link  link=${SEARCH_RESULT_TITLE}
     Page should contain  ${SEARCH_RESULT_TITLE}
 
-I can exclude content of type ${CONTENT_TYPE}
-    Unselect Checkbox  css=input[type="checkbox"][value="${CONTENT_TYPE}"]
-    Select From List By Value  css=select[name="created"]  today
-    Wait Until Element is Visible  css=dl.search-results[data-search-string*="created=today"]
-
```
mattss, line 87: Can I ask why this keyword was removed? It now means that the main robot test doesn't read like a user story. Abstracting the actual interactions into a keyword means that when the UI implementation changes, the keywords can be updated and the test cases can stay the same.

gyst:
@fredvd I agree with Matt. Can you please factor out the specifics into keywords in lib, then call those keywords?

mattss:
They keywords were already in place, so this should just be a case of reverting the 'simplify' commit.

fredvd:
@matts I remove the keyword because the extra filtering on created today items is what's probably causing the random failures and is not part of the test description. That left only one statement but I didn't know it isn't allowed to to explicit selects etc. in the main test descriptions. I've cleaned up the first keyword and added an extra line that checks for 'does not contain'. There's a similar new builtin keyword in selenium 1.7, but we're using 1.6 afaik.

mattss:
@fredvd Looks good - thanks!

mattss:
Test this please jenkins

Issue #573, opened 2015-07-17
Event: Add auto-suggest fields for Organiser and Invitees


cillianderoiste:
The Organiser and Invitees fields have not been enabled yet. P.a.events have fields for contact_name, contact_email, contact_phone which are also used for ical export. They don't have a field for Organiser, so rather than trying to use contact_name, or contact_email to save the user id of the Organiser, I think we should add a new field. If need be, we can save the Organiser name, email address and phone number to the contact_\* fields. The field can save a single user id, and display the user name.

I think the Attendees field can be used for Invitees, even though it has a slightly different connotation. This can save a comma separated string of user ids, and display the user names.

cillianderoiste:
Related to #431

cillianderoiste:
@pilz mentioned that we may eventually also want to indicate whether an invitee has confirmed or declined, so we might as well use a custom field for invitees too.

PR #605, opened 2015-07-17
Event: Add auto-suggest fields for organizer and invitees

cillianderoiste:
For #573


gyst:
These vocabularies should support a query filter so that only user suggestions matching the already-input search string are shown. Query matches should be made both against userid and user fullname, so that any user matches the query if either the username has a substring match or his full name has a substring match.

PR #606, opened 2015-07-17
check if there is a url in the post and link it

pilz:


pilz:
Is that a hint to review my own PR? :)

PR #617, opened 2015-07-20
Library search extensions

gyst:
work in progress


mattss:
@gyst got a bit further with this. 'phrase' is now optional if you supply at least one filter. The solr tests are failing for some reason because the content created in the subfolder doesn't seem to make it into solr correctly. Will continue to debug with @mgrbyte tomorrow.

gyst:
I've also seen Heisenerrors where some tests fail some of the time but pass other times, without code changes in between. This has led me to wonder whether there's possibly a delay in play where we're treating Solr as a realtime system when it isn't and it maybe hasn't properly updated its indexes when we already start querying.

pilz:
iirc you can send a flush to solr programmatically.

> Am 21.07.2015 um 20:52 schrieb Guido Stevens notifications@github.com:
> 
> I've also seen Heisenerrors where some tests fail some of the time but pass other times, without code changes in between. This has led me to wonder whether there's possibly a delay in play where we're treating Solr as a realtime system when it isn't and it maybe hasn't properly updated its indexes when we already start querying.
> 
> —
> Reply to this email directly or view it on GitHub https://github.com/ploneintranet/ploneintranet/pull/617#issuecomment-123440137.

## 

Alexander Pilz
SYSLAB.COM GmbH, Landwehrstrasse 60-62, 80336 Munich, Germany
Tel: +49 89 3063589-0 - Fax: +49 89 3063589-9
http://www.syslab.com - Amtsgericht Muenchen, HRB 135057
Steuernummer 143/184/50154 - Ust.-ID: DE212842815

mattss:
@gyst Issues now resolved.

I've simplified some of the solr test stuff. Previously we had registered various event handlers to force solr to update when you modified content in the tests, but I think this was causing too many solr updates and causing data oddities. Now reverted to just manually committing in the tests whenever changes are made that you want to see reflected in solr (due to the collective.indexing integration).

Path searching now works - see your test that now passes.

Also reworked the search docs slightly with some examples of interacting with search results alongside the APIs. Hopefully this clears things up a bit.

gyst:
Thanks a lot! Will check it out tomorrow.

Issue #451, opened 2015-07-08
unable to edit word file in workspace 


maartenkling:
http://localhost:8080/Plone/workspaces/open-market-committee/manage-information/minutes/view

how to edit the file? and replace it?

pilz:
We need to
- Enable the external editor feature (I propose by default)
- Probably should provide a download link to the client component somewhere
- Add a file widget to the file metadata to upload a new file

gyst:
These are two separate issues. I think externaleditor should NOT be enabled by default. That's an integrator's choice to enable, depending on the client IT landscape.

Which means we do need the extra upload widget to replace an existing file.
@cornae can you please add another icon to the meta bar?

gyst:
So this is about the upload icon.

External editor icon is discussed in
https://github.com/ploneintranet/ploneintranet.prototype/issues/77

cillianderoiste:
I'll add the file widget to files and images.

cillianderoiste:
I've added the fields in #626, I'll leave the rest to @cornae

PR #626, opened 2015-07-23
Replace file widget

cillianderoiste:
This adds file/image fields to files and images so that they can be replaced (in the metadata area). The fields still need to be designed, see #451.

The standard plone widget has an "action" field which determines whether the image should be replaced, removed or not changed. We only want to change the image when a new one has been uploaded. If we submit a form with an empty image field, the standard widget doesn't treat the empty field as a NOT_CHANGED value, so the image is deleted. In order to override the widget so that we can treat an empty string as NOT_CHANGED I replaced the model file, to remove the image/file fields and added an equivalent behavior instead. Then I overrode this widget for the IWorkspaceAppFormLayer. It seems like a super-awkward approach to achieve this, so if someone has a better idea, it would be great to hear about it.

I've verified that the file and image edit views still work correctly in the Library.

Since this change modifies the FTI for Image and File it requires re-importing the Types Tool. Should I add an upgrade step?


gyst:
Yes please add the upgrade, needed for iKath.

Issue #633, opened 2015-07-27
Search facets should support untagged (was: should be unchecked initially)


cillianderoiste:
With SOLR enabled:
1. Create 2 Documents: "Doc A" with the tag TagA and "Untagged Doc"
2. Search for "Doc"

The search facet for Tags will show "TagA" with the box checked.

If you uncheck the box, you still get two results. If you check it again you only get the document which has been tagged with TagA.

I guess that all facets should be unchecked initially.

cillianderoiste:
Hi @cornae, @mattss mentioned that the facet checkboxes are checked initially in the prototype and I see there is also a checkbox for "All other tags" which we haven't implemented (yet?). How should this work?

cornae:
The all other tags is a combination checkbox for all tags that are not in the list. So if it's ticked, all tags that are not in the checklist are shown. If unticked, those are hidden, but tags that are in the tag list might still be shown.

cillianderoiste:
@cornae thanks, I guess that means we only include the top 8 tags, and that "all other tags" also includes "no tags", and we'll need this option for any facet which could be empty. So in the example above:
1. Search for "Doc"
2. Both "TagA" and "All other tags" are checked
3. Results are "Doc A" and "Untagged Doc"
4. Click Tags->"None"
5. No results
6. Check "All other tags"
7. Get "Untagged Doc"
8. Check "TagA" and uncheck "All other tags"
9. Get "Doc A"

I see that the tags aren't ordered alphabetically, so I guess they are ordered by popularity in the search results (but probably are ordered alphabetically where they have the same amount of results).

Does that sound right?

It seems a bit odd to me to include untagged items in the results along with less popular tags, so I hope I've understood this correctly.

cornae:
The untagged one is an interesting one. I could also imagine that we add an extra box for 'untagged'. That could solve some issue. @gyst, How do you feel about that?

gyst:
Ok with me.

On August 10, 2015 2:08:02 PM GMT+02:00, "Cornelis G. A. Kolbach" notifications@github.com wrote:

> The untagged one is an interesting one. I could also imagine that we
> add an extra box for 'untagged'. That could solve some issue. @gyst,
> How do you feel about that?
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/ploneintranet/ploneintranet/issues/633#issuecomment-129423827

## 

Sent from my Android device with K-9 Mail. Please excuse my brevity.

[...]

PR #634, opened 2015-07-27
Uncheck search facets initially #633

cillianderoiste:


mattss:
I think this may need some discussion, especially as the prototype has them selected by default.

cillianderoiste:
Thanks, I've asked @cornae for feedback on #633.

gyst:
Following the discussion on #633 this PR is invalid.

Edit: fixed typo

PR #639, opened 2015-07-29
Fix unicode searches

cillianderoiste:
IIUC Solr expects the query to be a bytestring, and the spellchecker
expects a unicode string.

To reproduce the issue, search for "Zürich".


pilz:
deleted

pilz:
@bfackland  that might be the fix that I asked you for yesterday.

bfackland:
@mgrbyte It looks like the PI search unicode issue I mentioned earlier has been resolved already :)

mgrbyte:
@bfackland Awesome :hourglass: machine :)

PR #647, opened 2015-08-06
Fix tests.

mgrbyte:
@mattss Fixes test breakage as a result of #639 

~~(I think this was not picked up on due to no CI job using a solr.cfg?)~~
Update: Solr tests currently failing on jenkins


mgrbyte:
After speaking with @mattss , #639 (and this PR's adjustment) may be the wrong fix.
"The values are supplied as byte-strings because the values are pulled directly from the request.
We should use a z3c form to access the search form variables, in which case the values should already be decoded as unicode."

PR #656, opened 2015-08-09
Search results fragments

reinhardt:
When following a link in the search results the sidebar section corresponding to the portal_type is opened, e.g. if I click a Todo item, the todos in the sidebar are open.


reinhardt:
Similar to #479

reinhardt:
Hang on, I also need to check the other results templates, like file_results...

reinhardt:
OK here we go again.

PR #660, opened 2015-08-10
Solr index binary content

mattss:
Adds support for indexing of binary content using Apache Tika (built into solr).

Also fixes the solr tests which were broken by https://github.com/ploneintranet/ploneintranet/pull/639


mattss:
FYI @pilz

pysailor:
@matts Great! I'll have a look this week since pilz is on holidays for the next 2 weeks

pysailor:
Works really well! :+1:

PR #663, opened 2015-08-11
use correct way for string substitution in messageid

pysailor:
See http://docs.plone.org/develop/plone/i18n/internationalisation.html#translation-string-substitution


ale-rt:
Sorry @pysailor, do you want this merged? 
It tells me there are unmerged commit.
Anyway for me it's fine to merge.

Hard day today :) Making some mistakes

Issue #600, opened 2015-07-17
Event validation start date > end date


ale-rt:
At the moment it is allowed to create a start date greater than the end date.
This leads to the funny effect that the event is shown as a past event and a future event

gyst:
#481 failed to add front-end validation to prevent that. Which means we have to catch this in the backend. We can either refuse to save, if we can show the error in the UI, which has my preference. If this is not easy, let's at least then rewrite the end date to start+1hour so our backend is consistent. @ale-rt do you agree?

ale-rt:
+1

pilz:
Careful. Afaik we don't have a concept in showing backend errors properly in the frontend at the moment - but in a portal_statusmessage. This is not really nice. So before investing more than trivial time into this, the proposed approach in #481 using validation.js could be considered instead.

gyst:
Workaround: if end < start: force end to start + 1hour. Can be done in backend without requiring any UI change.

pilz:
+1

pilz:
JC has done a pattern for validate.js which is able to do this. We "just" need to integrate it now and use it instead of the parsley one.

ale-rt:
I am checking pat-validate but it does not seem to be easy to have an invariant validator like the one we need here.

We need to check  4 inputs:
1. a start input type=date
2. a start input type=time
3. an end input type=date
4. an end input type=time

the logic is:
if 1<3 -> ok
elif 1>3 -> ok
elif 1=3 and 2<4 ok
else ko

Any hints?

jcbrand:
@ale-rt pat-validate is old and based on Parsley which doesn't work well with the patterns approach. I created a new pattern pat-validation which does validation between dates. We're not using it in ploneintranet yet, but Syslab is already using it in another project. The idea being to start using it in PI as well.

Eventually we want to deprecate and remove pat-validate.

ale-rt:
Ah ok! I did not understand correctly previous comments!
To close this story I am implementing the workaround suggested in https://github.com/ploneintranet/ploneintranet/issues/600#issuecomment-122847127

ale-rt:
#668 and #671 should close this activity.

In case of any objection please merge at least #676, which modifies injection to make the status messages appear on the page after saving.

gyst:
#668 and #676 are merged, #671 ready to merge.

PR #670, opened 2015-08-12
Enable download

pilz:
Currently, files can't be downloaded anymore. This reenables the sharing icon on the document view with the only action to download the item.

### DIFF
--- a/src/ploneintranet/workspace/basecontent/baseviews.py
+++ b/src/ploneintranet/workspace/basecontent/baseviews.py
@@ -162,6 +162,21 @@ class ContentView(BrowserView):
                 return 'icon-file-{0}'.format(icon_name)
         return 'icon-file-code'
 
+    def content_type_name(self):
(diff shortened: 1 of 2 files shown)


pilz:
Module ploneintranet.workspace.browser.add_content, line 146, in default_end
  Module DateTime.DateTime, line 440, in **init**
  Module DateTime.DateTime, line 757, in _parse_args
TimeError: Invalid time: 2015/08/12 24:45

How is that possible? I try again hoping for a one timer

pilz:
jenkins please retest this

ale-rt:
Reegarding your issue the problem is here:
- https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/workspace/browser/add_content.py#L145

It should be:

``` python
str((int(parts[0]) + 1)%24)
```

or even better work with DateTime() + 1./24.

I am using a similar approach in #600

ale-rt:
Added #673 to quick fix this

PR #672, opened 2015-08-13
Do not reload page

ale-rt:
I removed a reload ecause if injection works properly we should just click on older events to see the creted event.
This implicitly tests the sidebar injection is working fine.


ale-rt:
Jenkins test this please

PR #673, opened 2015-08-13
fix default_end method

ale-rt:
It was failing from 23 to midnight

### DIFF
--- a/src/ploneintranet/workspace/browser/add_content.py
+++ b/src/ploneintranet/workspace/browser/add_content.py
@@ -142,7 +142,7 @@ class AddEvent(AddContent):
         date = now.Date()
         time = self.round_minutes(now.TimeMinutes())
         parts = time.split(":")
-        parts[0] = str(int(parts[0]) + 1)
+        parts[0] = str((int(parts[0]) + 1) % 24)
         result = DateTime(date + " " + parts[0] + ":" + parts[1])
         return result
 


pilz:
@ale-rt you are great!!

ale-rt:
Come on, it was a one liner! :)

pysailor:
@ale-rt Nice, but I could have sworn I had fixed that too last weekend... Ah, it was in a branch where I noticed that jenkins had failed shortly before midnight because of that... And I never pushed that

In my change, I also incremented the date by 1 in case the hour was the one before midnight. here's my diff:

```
diff --git a/src/ploneintranet/workspace/browser/add_content.py b/src/ploneintranet/workspace/browser/add_content.py
index a5bc56b..771d7f4 100644
--- a/src/ploneintranet/workspace/browser/add_content.py
+++ b/src/ploneintranet/workspace/browser/add_content.py
@@ -139,11 +139,14 @@ class AddEvent(AddContent):

     def default_end(self):
         now = DateTime()
-        date = now.Date()
         time = self.round_minutes(now.TimeMinutes())
         parts = time.split(":")
-        parts[0] = str(int(parts[0]) + 1)
-        result = DateTime(date + " " + parts[0] + ":" + parts[1])
+        hour = int(parts[0])
+        hour_end = (hour + 1) % 24
+        if hour_end < hour:
+            now = now + 1
+        date = now.Date()
+        result = DateTime(date + " " + str(hour_end) + ":" + parts[1])
         return result

     def round_minutes(self, time):
```

ale-rt:
Agree, I was just preventing other failures.

In my opinion everything can be fixed in an easier and more significant way by defining:

```
nowplusone = DateTime()+1./24
```

instead of now!

pysailor:
:+1:

PR #674, opened 2015-08-13
Fix double close button in delete popup

ale-rt:
Fixes double cancel button when clicking on the trash icon in the events sidebar:

image


ale-rt:
To the reviewers: please before merging this give priority to #675!

PR #698, opened 2015-08-19
Supervisor

pilz:
This is an initial supervisor config depending on zeo and solr and integrates celery support. However none of the celery tasks are really hooked up yet. 
Also this requires you to have redis installed on your system.

You need to run complete.cfg instead of coredev or buildout. I kept this separate for now until we are really happy and confident that we want to switch this to the default.


pilz:
test please

PR #699, opened 2015-08-19
Reply security

gyst:
This fixes the following scenario:
- Workspace member Alex posts a statusupdate in the workspace
- Workspace member Bernard replies to this update in the global stream. He can view it in the stream because he is a workspace member.
- Non-workspace-member Clare loads the dashboard. She should see neither the original post nor the reply.

The bug was that until now the reply was rendered, which then broke on Unauthorized (yay! for our security!).

The fix is to make replies to a toplevel post inherit the security context of that toplevel post.
This fix includes a database migration to also fix existing older replies.
As part of this PR a bit of test refactoring and cleanup was needed.

A robot test for the scenario is provided.

### DIFF
--- a/src/ploneintranet/microblog/configure.zcml
+++ b/src/ploneintranet/microblog/configure.zcml
@@ -47,7 +47,6 @@
     source="0001"
     destination="0002"
     handler=".migration.setup_uuid_mapping"
-    sortkey="1"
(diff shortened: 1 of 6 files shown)


gyst:
@pysailor this one is for you.

pysailor:
Looking good, thanks!

PR #706, opened 2015-08-20
Todo: select assignee from all users, not just WS members

cillianderoiste:


mattss:
Why would you assign a todo within a workspace to someone who wasn't a member of a workspace? Raises some permissions questions for me.

cillianderoiste:
Sorry for not mentioning the use case, and having re-read the ticket I see it's not complete yet. I'll try to get clarification but IIUC @pilz and @gyst agree that it should be possible to assign users to tasks, invite them to events etc. even if they can't view the object (Redmine 12219).

gyst:
On 20/08/15 18:17, goibhniu wrote:

> Sorry for not mentioning the use case, and having re-read the ticket I
> see it's not complete yet. I'll try to get clarification but IIUC @pilz
> https://github.com/pilz and @gyst https://github.com/gyst agree that
> it should be possible to assign users to tasks, invite them to events
> etc. even if they can't view the object (Redmine 12219).

It's more complex.
- Users should be assignable to tasks: OK.
- Users should be invitable to events: yes and also gain view permission
  on the event, though not on it's container.
- "etc" is quite constrained to specific metadata fields on a Case that
  have no further UX or security impact. Please ask @pilz what fields
  exactly need this enlarged user set.

In no case should this be applied as a blanket change to all user pickers.

## :*CU#

```

```

pilz:
I see a strong use case to assign users outside of a workspace to a task to keep track on who has to still perform certain tasks. Not everything goes on in a workspace, and many real world tasks might still need to be completed by others which wouldn't want to be bothered with yet another workspace access just becasue they have to get back with an info to the team. Example: Need ok from general vicar. Somebody needs to chase him, not necessarily one single person, but the whole team. So the important info is that the vicar hasn't yet delivered. And he doesn't want and perhaps even shouldn't get access to the WS.

Currently we leave it to the responsibility of the user to add the necessary users to a workspace so that tasks can be assigned. In the future we still leave it to the users to make sure people have access to these tasks, but enforce that a bit later (in a lazy way). If you are adding some tasks, everytime one of the users is not in the WS, it kills the flow of recording these activities. 

Similar to the events, we may think about automatically adding localrole edit rights to a task once a user is assigned.

gyst:
Is already in ikath. Risk: this can expose tasks on a user's dash that the user has no access to.

Issue #691, opened 2015-08-19
solr PR builder


gyst:

mattss:
Unfortunately, Jenkins-github integration is limited if you have multiple PR checkers. It seems that the first check that completes will update the PR with 'everything's fine', even if the other one is still running.

The alternative approaches I see therefore are:
1. We continue to run the solr search tests manually on master pushes (simplest, but with obvious downsides)
2. We update the existing PR job to run the solr build and solr-search tests _as well_, after the main tests. This is the next simplest (adding a few lines to jenkins configs) but will require jenkins to build out multiple times and considerably lengthen the test times.
3. We ditch the zcatalog-based search, make solr a requirement and run all tests with the solr config every time. This is the most complicated, as it involves merging the buildouts and possibly adding supervisor, and updating all the relevant documentation.

Personally I would vote for 1 or 2 now (in the interest of getting venus out), and then look into 3 for the next release.

Any thoughts?

pilz:
I fully agree. 
I would even say that as long solr is not fully integrated as the default, it is also not the priority and we should do 1.
Once we integrate solr via super isor as default, it superceeds zcatalog as the default and we do 3. after venus. 

Alexander Pilz (mobil)

pysailor:
I would also favor 1 or 2 for now. Considering that the tests for PRs already take a long time to complete, I opt for 1.

gyst:
On 21/08/15 11:33, Matthew Sital-Singh wrote:

> Unfortunately, Jenkins-github integration is limited if you have
> multiple PR checkers. It seems that the first check that completes will
> update the PR with 'everything's fine', even if the other one is still
> running.
> 
> The alternative approaches I see therefore are:
> 1. We continue to run the solr search tests manually on master pushes
>    (simplest, but with obvious downsides)
> 2. We update the existing PR job to run the solr build and solr-search
>    tests /as well/, after the main tests. This is the next simplest
>    (adding a few lines to jenkins configs) but will require jenkins to
>    build out multiple times and considerably lengthen the test times.
> 3. We ditch the zcatalog-based search, make solr a requirement and run
>    all tests with the solr config every time. This is the most
>    complicated, as it involves merging the buildouts and possibly
>    adding supervisor, and updating all the relevant documentation.
> 
> Personally I would vote for 1 or 2 now (in the interest of getting venus
> out), and then look into 3 for the next release.
1. We update the existing PR job to /only/ run with solr.

[...]

PR #711, opened 2015-08-21
Jenkins now builds using solr config

mattss:
Resolves https://github.com/ploneintranet/ploneintranet/issues/691


mattss:
Mmmmm robot test isolation issues. There's my afternoon sorted.

PR #718, opened 2015-08-24
Enable autoexpand on scroll down

pilz:
This loads more stream when scrolling to the end.

@gyst my head is dingling from reading through microblog ;) and I am not sure I am on the right track. Could you take a look whether what I do is sane and in the right place? It seems to work, but did I miss something?

### DIFF
--- a/src/ploneintranet/activitystream/browser/tiles/stream.py
+++ b/src/ploneintranet/activitystream/browser/tiles/stream.py
@@ -19,8 +19,6 @@ class StreamTile(Tile):
 
     index = ViewPageTemplateFile("templates/stream_tile.pt")
     count = 15
-    b_start = 0
-    last_one = False
 
     def __init__(self, context, request):
         self.context = context
@@ -30,6 +28,9 @@ class StreamTile(Tile):
         self.explore = 'network' not in self.data
         if 'b_start' in request:
             self.b_start = int(request.get('b_start'))
+        else:
+            self.b_start = 0
+        self.last_one = False
 
     @property
     def b_next(self):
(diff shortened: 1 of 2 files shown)


gyst:
I've pushed a small change.

But more fundamentally the chosen implementation is not the most performant algorithm possible here. You're now fetching increasingly larger batches by upping `limit` until you've exhausted the whole microblog key space. This results in a sequence of database loads where each consecutive load fetches all items the previous load already fetched, plus ten more, and then uses only the last 10.

Instead of using `b_start` a more performant algorithm would use the id of the last (oldest) update shown (propagated via the template and request), and use that as a `max` argument on the microblog accessor, with a `limit=10` batch size every time. This would avoid fetching the already rendered statusupdates. Id is a creation microtime longint (millisec since epoch). To avoid loading the last-rendered update again in the next batch, the query should be `max=long(id)-1` which reads as _all updates with a creation up until, but not including, `id`_.

pilz:
Jep. That is where my head started dingling. But then I thought, this is accessed linearly anyway, so all previous items are in the cache already and thus loading is cheap. But then this is not true for a zeo setup again…
I guess this needs some more understanding of the underlying data structure…

gyst:
It's a microtime indexed btree (think: persistent dict). Key operations are very very fast in C so giving the accessor the right `max=id-1` and `limit=10` results in _real_ linear access, fetching 10 non-overlapping updates per query. I'm happy to do that together with you tomorrow as part of the spike, shouldn't be that hard.

pilz:
perfect

gyst:
Hi Alex, I refactored your work. Re-assigning to you for final review.

pilz:
Ah, that easy? Frustrating :)

PR #724, opened 2015-08-26
Task: Display the "planned" workflow state

cillianderoiste:
There are three possible workflow states for a Task: "open", "planned"
and "done".

An "open" Task is assigned to the current Milestone of a Case and
appears on the dashboard. A "planned" Task is assigned to a Case
Milestone in the future and isn't listed on the dashboard.

Whether it is "open" or "planned" depends on the workflow state of the
Case and the Milestone of the task. It is updated by an event handler,
so it doesn't make sense to allow a user to manually switch between
"open" and "planned". A user can only "close" or "reopen" a task.

SLC Refs #11236


pilz:
test please

pilz:
test please

PR #725, opened 2015-08-26
Library utf8 tags

gyst:
Library tagging views were broken in solr. Now they're broken in zcatalog :-).

In a follow-up PR I will:
- deprecate the zcatalog tag views
- add robot test coverage on solr-backed library


gyst:
Please hold I'm hunting futher unicode bugs

pilz:
test please

src/ploneintranet/library/browser/templates/library.html
```diff
@@ -158,6 +159,11 @@ <h3 tal:content="child/title">page title</h3>
 
         </div>
 
+        <p tal:condition="not:view/enabled">
```
pilz, line 162: @gyst When calling the library, enable is not present and so the view gives a 500. Either a tal:on-error="nothing" or have the enabled everywhere. What do you think?

gyst, line 162: On 27/08/15 02:03, Alexander Pilz wrote:

> @gyst https://github.com/gyst When calling the library, enable is not
> present and so the view gives a 500. Either a tal:on-error="nothing" or
> have the enabled everywhere. What do you think?

Ah, stupid. I've set enabled=True on everything by default.

## 

```

```

PR #732, opened 2015-08-27
Search results: translate friendly_type_name

cillianderoiste:
We index the mime type name for files e.g. "Word 2007 document" and
for plone content types we use the name of the content type. Although
Products.Mimetypes has a registry where we can look up the relevant
translation for each file type it's tricky to do that for the search
facets since we don't index the id.

As a work-around, this uses the ploneintranet i18n domain and adds
msgids for the various content types and file types.


pilz:
test please

pilz:
test please

PR #733, opened 2015-08-27
shorten titles to 80 chars to cater for crazy germans

pilz:
If somebody adds insanely long titles (like germans love to do), the title first hides the searchbar and then , when even longer, breaks and is totally hidden. This crops it to decent 80 chars.
Also the workspace overview constrains on 80 in display, however if users like, they can still add as many chars as they want in the edit form. We don't want to be so picky, do we?

### DIFF
--- a/src/ploneintranet/layout/browser/overrides/plone.app.layout.viewlets.path_bar.pt
+++ b/src/ploneintranet/layout/browser/overrides/plone.app.layout.viewlets.path_bar.pt
@@ -1,6 +1,6 @@
 <nav id="breadcrumbs" class="breadcrumbs" role="navigation"
          i18n:domain="plone"
-         tal:define="breadcrumbs view/breadcrumbs">
+         tal:define="breadcrumbs view/breadcrumbs; ploneview here/@@plone">
 
         <!--a i18n:translate="tabs_home"
             tal:attributes="href view/navigation_root_url">Home</a-->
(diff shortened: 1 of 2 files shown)


cillianderoiste:
Why not do it in CSS? I haven't checked but cropping the characters server side can also cause problems for different window sizes, especially since some characters are wider than others.

pilz:
That was something I could do right away and it seems rather uncritical as we are not aiming at an exactl size. But I am fine with a css solution as well.

pilz:
Added an alternative solution in css to proto in https://github.com/ploneintranet/ploneintranet.prototype/pull/142. If @cornae accepts that, we can reject this here.

gyst:
Test failures appear to be unrelated to the diff. I prefer the css-only solution though, it's a rendering issue not a backend one.

pilz:
Yep. We just can't easily do the cut on a multiline workspace tile. the css ellipsis only works on one liners (or you need js again). I prefer the css solution plus a hard cut on the tiles.

pilz:
@pbauer is looking into the testfailure which seems to be an isolation issue that came in with the event validation.

cornae:
The problem is not in the output but in the input. Such long titles should never be allowed. If a user wants to add parts of the bible, the European constitution or the collected works Goethe, they can use the description for it. We should put a max char limit of 80 chars or less on the input field for workspace title input.

pilz:
<input type="text" name="title" maxlength="70"> ?

pilz:
input type="text" name="usrname" maxlength="10" ?

cornae:
YEs, but then for Workspace title, not user name. And something between 50 and 80 chars. We have to test what works well.

PR #736, opened 2015-08-28
upgrade to Plone 5.0b4

pbauer:


pbauer:
test please

pbauer:
The failing robot-test is unrelated since is also fails in other pull-requests. It might be due to the fact that events in the query `ploneintranet.workspace.browser.tiles.sidebar.Sidebar.events` have no `sort_on`. What is the wanted behavior? Older or newer dates on the top?

pilz:
Good catch. Many tests are running into this at the moment. The newer dates should be on top.

PR #741, opened 2015-08-29
Async dispatch

gyst:
This introduces the async browser-view-via-celery call flow we've been talking about since Sorrento.
It addresses authentication and CSRF protection.

This PR provides the async framework. Todo once it is merged:
- activate async as part of the suite dependencies
- hook up the reindexer
- hook up preview generation and cleanup ploneintranet.docconv.client.async
- replace `collective.celery` dependency with `celery` dependency only


gyst:
Oh, check out http://localhost:8080/@@async-status for fun. You'll need to be logged in as portal manager.

gyst:
@pilz Ready for review and merge.
@wengole @simonedeponti Would appreciate your review of this also.

simonedeponti:
@gyst sorry for the delay, but I gave a quick look and seems okay to me. Good work on the view protection code.

There is one thing that is not clear (from the code) to me, and it's whether or not the task is queued immediately or when the transaction commits. This is important to avoid race conditions (http://celery.readthedocs.org/en/latest/userguide/tasks.html#database-transactions).

Also, although not critical, I would provide a standard way for tasks to avoid double-queueing (this might be done on the application layer too, e.g. by annotation on the ZODB "this task has already been queued"). The case I have in mind is one where you (might) decide to generate the preview upon viewing, instead of when the upload takes place. This is however a tricky topic, as some instances of double-queueing (I can think of reindex) are actually legitimate, while others might not be . So, I would think about it but you can also decide not to act upon it immediately.

siebo:
I just gave things a quick look and haven't dug too deep in the code yet, but I think this is an awesome idea and glad you've made it a reality!

PR #747, opened 2015-09-01
Dashboard theme update

cillianderoiste:


pilz:
@cillianderoiste Can you check the failures? I bet this is due to changed selectors only.

PR #756, opened 2015-09-07
Use App.config to override themeswitcher settings

cillianderoiste:
This is copied from
https://github.com/collective/collective.xmpp.core/blob/master/collective/xmpp/core/dbconfig.py

It allows the themeswitcher hostname_switchlist to be configured from
buildout, which helps for deployment e.g. using batou.


ale-rt:
I learned a wonderful thing with this commit!
Did not know it was possible to register an event for IProcessStarting.

Am I wrong or this event will rewrite eventual registry modification every time the process starts?

PR #517, opened 2015-07-14
MetroMap: check transition guards for [close milestone]

cillianderoiste:
NOTE: this also sets manager_bypass to False. It's a bit confusing if
the manager can close milestones but other users can't IMO.

Issue #462, opened 2015-07-09
case templates bootstrapping


gyst:
This follows up on #292 

The main problem is, that this now hard requires templates to be configured, but:
- no templates are available in a default install (only in testcontent)
- there's no (documented) way to create a template TTW

In short, if we follow Maarten's procedure, and install an empty Plone Intranet site, this results in a broken product. 
1. We need documentation on how to configure templates from scratch.
2. Currently you cannot create a template TTW without a template...
   
   (a.) either we make it possible to bootstrap TTW, e.g. with sane defaults
   
   (b.) or we hide all "Cases" functionality unless templates are available (from a custom setup handler)

My guess is that 2a. is easier than 2b. It's also more elegant since 2b. requires integrators to write custom code before they can even use this part of the product.

gyst:
Sane defaults could be implemented by changing the conditional here:
https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/workspace/subscribers.py#L62

cillianderoiste:
Thanks for the review!

I guess another option is to create and configure an example Template Case in the workspace setuphandlers: https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/workspace/setuphandlers.py#L30

Let me know which option you prefer and I can implement it and add some documentation for integrators which describe how to create additional Case Templates and configure them (via the ZMI). Does that sound OK?

I do wonder if there is a plan in the future to create a user interface for managing cases. Perhaps it would also be a powerful feature for intranet admins.

gyst:
@cillianderoiste I'd like a hardcoded fallback template to be available at all times. Templates living in content space are easily damaged by content editors; providing a code level fallback makes the system resilient and allows user to recover from errors.

cillianderoiste:
OK, that sounds good.

gyst:
Fixed in #773

PR #758, opened 2015-09-11
update_todos_state: perform transition as temporary Editor

cillianderoiste:
This allows Todo transitions to be protected by the Editor role, and
also for the workflow state to be set to Open or Planned as
appropriate.

Without this change it is possible for a user to perform a transition on
a Case which puts it into a state where that user no longer has the
permissions to edit Todo items. The event handler will then try to
update the workflow state of Todo items e.g. from Planned to Open and
this will fail, which will cause the whole transaction to fail.


cillianderoiste:
retest this please

gyst:
Putting this PR on hold since it has no test coverage, and testing it manually breaks

Module ploneintranet.workspace.adapters, line 292, in metromap_sequence
NameError: global name 'available_transition_ids' is not defined

@cillianderoiste Can you please NOT disable your pre-commit hooks?

/app/src/ploneintranet/workspace/adapters.py:292:45: F821 undefined name 'available_transition_ids'

gyst:
I will work on #758 and #517 and #462 today.

cillianderoiste:
@gyst sorry about the code-analysis failures, the pre-commit hooks never worked for me ... I must fix that.

gyst:
Superseded by #773, closing.

Issue #731, opened 2015-08-27
jenkins is unstable


gyst:
When triggering pull request builds I frequently get broken builds:
./bootstrap_jenkins.sh: line 16: ./bin/buildout: No such file or directory

The builds that do run then frequently hang at the start of the robot testing

in 19.122 seconds.
  Set up plone.app.robotframework.remote.AutoLoginRemoteLibrary:RobotRemote in 0.013 seconds.
  Set up ploneintranet.suite.testing.PLONEINTRANET_SUITE_ROBOT in 0.000 seconds.
  Running:
.The following test left new threads behind:
Page contains just one meta for viewport (test_transforms.robot)
New thread(s): [<_DummyThread(Dummy-23, started daemon 140714358474496)>]

(loading icon keeps circling for ages)

After a while these are then disappear, then re-appear, in the Jenkins console (as if they're auto-restarted by Jenkins) but the console output remains hanging, though some come through in the end and mark a PR green.

PR ended up green: 
https://github.com/ploneintranet/ploneintranet/pull/729

PRs that keep hanging in orange:
https://github.com/ploneintranet/ploneintranet/pull/683
https://github.com/ploneintranet/ploneintranet/pull/681
https://github.com/ploneintranet/ploneintranet/pull/671

This sad state of affairs obviously makes it nigh impossible to clear out the PR backlog and get code into master. @mattss I hope you can find a solution.

gyst:
Additionally, we now have the master builds failing for reasons that are independent of code changes on merges that were pre-tested green on the PR builder.

http://jenkins.ploneintranet.net/job/complete.Ploneintranet.master.False/414/
http://jenkins.ploneintranet.net/job/complete.Ploneintranet.master.False/415/

415 for example was a no-op change versus 414 - only documentation update.

mattss:
@gyst can you get the master build to pass locally?

gyst:
On 27/08/15 14:28, Matthew Sital-Singh wrote:

> @gyst https://github.com/gyst can you get the master build to pass
> locally?

Will run that now.

## 

```
Guido Stevens  |  +31.43.3618933  |  http://cosent.nl

s o c i a l   k n o w l e d g e   t e c h n o l o g y
```

mattss:
I _think_ this may be because the pull request builder will often run multiple builds at once, in the same workspace.

The simple fix would be to make sure it only runs one at a time, which would obviously reeaally slow things down during a sprint.

[...]

PR #762, opened 2015-09-14
Simplify docconv

pilz:
Hold, this is only to abuse jenkins for a test run. Do not merge!


pilz:
@gyst @wengole Here is a version that rips out nearly all of docconv. This now is using as much of collective.documentviewer as possible.

I followed Ben's approach to move necessary access code to pi_api and moved also the tests for it there. Now, docconv.client contains also very minimal code to setup and config c.dv, to register our own handler and to provide the indexer that let's us know from the catalog if there are previews available. I bet these can go into other packages and we can remove docconv.client for good.

I would be very happy about a review so that I can start to make the classic workspace upload async.

gyst:
@pilz can you merge master and resolve the conflicts? Would like to review and merge this.

pilz:
done

src/ploneintranet/activitystream/browser/prototype/post.html
```diff
@@ -59,13 +59,34 @@
...
         <img src="/media/preview_thumb_1.jpg" alt=""
              tal:attributes="src string:${attachment/img_src}">
       </a>
+          <figcaption>
+            <strong class="title">
+              <a tal:attributes="href attachment/link; alt attachment/alt" href="/projection-material" alt="Projection Material">
+                <tal:r replace="attachment/title">Projection Material</tal:r> <!--span class="page-counter icon-docs">3x</span-->
+              </a>
+              <!-- This needs exposure of all generated thumbs. We can do that and have them accessible, it is just not added here yet. -->
+              <!--span class="functions">
```
gyst, line 75: Why is this not exposed, and what needs to be done for that? Can you maybe add a ticket for that?

pilz, line 75: Cornelis has done additional design for this and I got that in while updating the prototype. I took over the full markup but to implement it, we need to do some tweaks to get more than one preview in. Not difficult, but another "Nebenkriegsschauplatz". At the moment it behaves as before. I consider that a new feature. 

I don't feel qualified to add that ticket as I have not initiated that design change, I frankly don't know more than what I can see from the markup.

gyst, line 75: OK he talked me through that the other day so I guess this is on my plate ;-).

gyst:
Apart from the fact that this change is too big for me to comprehend (my problem), a few remarks:
- I really really dislike doing actual implementations in ploneintranet.api. If we follow this route we can put everything in ploneintranet.api and don't need any other namespaces anymore. Doing the actual preview logic (and testing those) belongs in ploneintranet.attachments IMO. Especially a massive private function like `_get_dv_data(obj)` destroys the readability of the API.
- It's better to have theme/generated changes in separate pull requests, but of course not a show stopper.

pilz:
Ok, so we could actually move this back into docconv.client and just reference it from the api. I don't think that would be a big issue.

pilz:
jenkins please test this

gyst:
PR build fails because of #731.

pilz:
test please test

pilz:
test please test

pilz:
some obscure github checkout problems are happening

pilz:
@gyst review now while it is still hot :)

gyst:
@pilz waiting for your promised refactorings...

pilz:
ah, ok

pilz:
@gyst the preview code is back in docconv.client now, the api does only call it.

PR #768, opened 2015-09-15
build complete stack by default

gyst:
Our stack requires celery and solr. Build that by default, and ZEO as well.


gyst:
Doesn't impact jenkins.cfg so jenkins testing this is rather irrelevant.

gyst:
Please ignore test failure, irrelevant for this PR.

PR #770, opened 2015-09-15
don't link the word profiles to the profiles folder

### LINKED ISSUE #755: Proposal to list all users when no search term is given by default
A quick solution to our problem that the profile folder has no proper view could be to show the people_search form. But then it should list users by default. It already does batching - showing only 15 users by default - and would be a great entry point for searching more. The only change required would be not to abort if there are no search keywords. 

@mattss, @gyst: Does that make sense?

pilz:
Fixes #755 temporarily until we have a better solution.


pilz:
xvfb error in jenkins. please test please

gyst:
@pilz can you review my changes and merge?

pilz:
Test breakage is correct. Needs to be rewritten to test for breadcrumbs-current now.

gyst:
Will fix

PR #772, opened 2015-09-16
Use pat-validation instead of pat-validate

jcbrand:
I've gone through the page templates and changed all occurrences that use the old deprecated pattern pat-validate with the new and shiny pat-validation.

This means that we now have before/after date validation on the events page. Unfortunately only for the date field, not the date and time fields together. Handling both would be a more tricky because two different inputs are being used there. Not yet sure how to approach that usecase.

cc @pysailor, @pilz


gyst:
Multiple test failures.

jcbrand:
I'm not able to reproduce the last  3 errors from Jenkins locally.

PR #777, opened 2015-09-18
AllUsersJSONView: use the catalog to search for users

cillianderoiste:
Since we have userprofile objects, we can avoid expensive PAS/LDAP
lookups.  Even on a local instance without LDAP the time to query 50
users went from 1.5 seconds to 40ms.


gyst:
@cillianderoiste Can you verify that this still works as intended after my refactoring?

cillianderoiste:
I've tested your changes @gyst and it's working great. I also added an extra change so that we can get results for partial matches.

PR #778, opened 2015-09-18
Jenkins should depend on the complete config

mattss:
Fixes the recursive dependencies introduced by https://github.com/ploneintranet/ploneintranet/pull/768


mattss:
Ditching this as it's not enough, and @pysailor is on the case.

PR #780, opened 2015-09-21
Create Workspace: avoid acl_users caching issues

cillianderoiste:
When caching is enabled on acl_users, creating a new workspace fails.

This bypasses security checks for transitioning a freshly created
workspace to the correct state and then invalidates the cache so that
the user who created the workspace has the appropriate roles on the
workspace.

NOTE: this certainly needs cleanup, but it should be enough to allow us
to cache acl_users.

This modifies ploneintranet_workspace_workflow so that the workspace
Owner has "Access contents information" in the default workflow
state. To test, you will need to import the workflow, or apply the
change manually.


src/ploneintranet/workspace/browser/add_content.py
```diff
@@ -92,10 +96,25 @@ def create(self):
...
                     raise AttributeError
-
+            old_security_manager = getSecurityManager()
+            acl_users = api.portal.get_tool('acl_users')
+            tmp_user = UnrestrictedUser(
+                'Workspace Init', '', ['TeamManager'], '')
+            tmp_acl_user = tmp_user.__of__(acl_users)
+            newSecurityManager(None, tmp_acl_user)
+            try:
                 new.set_external_visibility(external_visibility)
```
gyst: File "/var/lib/jenkins/workspace/Plone.Intranet.Pull.Request.Builder/src/ploneintranet/workspace/browser/add_content.py", line 106, in create
    new.set_external_visibility(external_visibility)
UnboundLocalError: local variable 'external_visibility' referenced before assignment

This is because upwards (not shown in the diff) there's no fallback clause to define this var if there is no 'form'.

gyst:
@cillianderoiste This does two things: 1. perform an operation with escalated privileges (set the security policy) and 2. invalidate the cache (so the user gets to use the newly set policy). 

You might want to refactor (1) to use the new `unrestricted.execute_as_manager` function I'm introducing in https://github.com/ploneintranet/ploneintranet/commit/9dad9121c1f289c5c7f8dfdf544e0b503849cff9 - better to not sprinkle security privilege escalation implementations all across our codebase. (I will refactor the todo transitions in #758 to use `unrestricted` myself.)

gyst:
PS to get the `unrestricted` module you can fork off branch `case-refactor` for now. I hope to finish that work soon.

cillianderoiste:
Beautiful, thanks!

Issue #33, opened 2015-03-13
Roster membership is cached (for an hour?)


wengole:
<a href="https://github.com/fredvd"><img src="https://avatars.githubusercontent.com/u/394784?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [fredvd](https://github.com/fredvd)**
_Tuesday Aug 05, 2014 at 12:36 GMT_
_Originally opened as https://github.com/ploneintranet/ploneintranet.workspace/issues/29_

---

It's not really a bug but a feature coming from collective.workspace. The PAS group plugin for the workspace imports the OFS.Cache Cacheable mixin which I assume has the result that when you start to play with a demo workspace and add/remove some users, the membership is cached. Default cache time for the RamCache in Plone is 3600 seconds. I first thought there was something broken in the workspace code, but a simple invalidate for acl_users in the RamCache helps. 

For now it would be cool to add some documentation for this behaviour if this is intentiional and not something that could be fixed in the short run. Maybe it's possible to flush the pas plugin cache when users are added/removed, but when you have a lot of workspaces/users that could be a huge penalty when everything is flushed.

wengole:
<a href="https://github.com/mattss"><img src="https://avatars.githubusercontent.com/u/1155400?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [mattss](https://github.com/mattss)**
_Wednesday Aug 13, 2014 at 08:19 GMT_

---

This surprises me, as it is not something we encountered during development or testing. Can you outline the steps required to replicate it, or put together a test to do so?

Many thanks.

wengole:
<a href="https://github.com/wengole"><img src="https://avatars.githubusercontent.com/u/383857?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [wengole](https://github.com/wengole)**
_Wednesday Mar 11, 2015 at 18:28 GMT_

---

Unable to reproduce

wengole:
<a href="https://github.com/fredvd"><img src="https://avatars.githubusercontent.com/u/394784?v=3" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [fredvd](https://github.com/fredvd)**
_Thursday Mar 12, 2015 at 10:06 GMT_

---

[...]

PR #781, opened 2015-09-22
Fire an event when the workspace roster is changed

cillianderoiste:
This can be useful for clearing caches, related to #780.


gyst:
Ref #33

pilz:
please test please

PR #785, opened 2015-09-23
hide all our components except suite:default

gyst:
Unclutter the add-ons control panel.


pbauer:
I'd prefer to have two lists that you can combine:
One with all the default-profiles and one with all the uninstall- and demo-profiles. That would make it a little easier to modify and see what's there.

gyst:
Why? I've made sure to sort the list alphabetically for easy access. We don't distinguish between default and other profiles; they should ALL be hidden.

PR #787, opened 2015-09-23
Case: fix case metromap avatars

cillianderoiste:
The workspacefolder.tasks() method was changed to return User objects,
for the sidebar metadata. This broke the avatar images in the
metromap. Since we already have the User objects, we don't need to call
existing_users_by_id to get the details (which is probably quite
expensive).

TODO: test both the metadata in the sidebar and the metromap to prevent
similar regressions in the future.

SLC Refs #12393


gyst:
Failures are webdriver heisenbugs, not real bugs.
Jenkins please test this test please.

gyst:
Jenkins please test this test please

PR #773, opened 2015-09-16
Case refactor

gyst:
This PR makes cases in Plone Intranet releasable for Venus. Fixes:
- provide a properly secured workflow
- provide a good demo case
- provide a good default case template in the default install (fixes #462)
- make it possible for non-admins to create a case from a template
- switch todos from 'planned' to 'open' even on protected milestone transitions (replaces #758)
- activate workflow guard on forward state transitions (replaces #517)
- sanitize metromap logic
- provide test coverage on the above

PR #798, opened 2015-09-28
Task sidebar: show task metadata for Workspaces

cillianderoiste:


gyst:
This is a one-line change but because of code reindents explodes to a 20-line diff that will conflict with the work I've done in #773.

gyst:
Trying to apply the metal call on top of #773 breaks the workspace completely. I also don't see any such "additional-meta-data" metal macro defined anywhere. PR invalid, closing.

pilz:
@cillianderoiste Can you please check what went wrong here?

PR #816, opened 2015-10-02
815 stream markup

gyst:
Fix the stream so stream attachments are only linked via the download icon, not the title.
Also show uploaded images in full glory.

If somebody can get the 'large' scale to work that would speed up user performance.

### DIFF
--- a/src/ploneintranet/activitystream/browser/statusupdate.py
+++ b/src/ploneintranet/activitystream/browser/statusupdate.py
@@ -193,9 +193,11 @@ class StatusUpdateView(BrowserView):
             self.attachment_base_url,
             item.getId(),
         ))
+        is_image = False
         if pi_api.previews.get(item):
             url = '/'.join((item_url, 'small'))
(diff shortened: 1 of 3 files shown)


src/ploneintranet/activitystream/browser/statusupdate.py
```diff
@@ -193,26 +193,29 @@ def item2attachments(self, item):
...
+            #     'images',
+            #     item.aq_base,
+            #     self.request,
+            # )
+            # url = '/'.join((
+            #     item_url,
+            #     images.scale(scale='large').url.lstrip('/'),
+            # ))
+            # # use unscaled instead
+            url = item_url
```
gyst, line 211: @pilz Any clue why the 'large' scale would return a 'preview' size image?

pilz:
@gyst not really. the large scale in plone indeed is set to 768 and c.dv uses 1000, 700 and 180 as sizes, so it is unlikely to come from there.

PR #820, opened 2015-10-07
Task sidebar updates metromap 2

cillianderoiste:
This is a copy of #742 which was then reverted in #748.

To reproduce the problem:

Open a Case
Open the Tasks tab in the sidebar
Open a milestone e.g. Prepare
Click the [Create new task] button within that milestone

The modal which is displayed, looks wrong:
metromap-injection

And when you click [Create] two tasks are created.

Interestingly, I don't get this problem with my iKath instance, so perhaps my JS bundle is out of date.


jcbrand:
@cillianderoiste Working now.

pilz:
Big Kudos for this one @jcbrand !!

PR #821, opened 2015-10-07
Userprofile fixes

### LINKED ISSUE #779: Update passwords to bcrypt
https://pypi.python.org/pypi/bcrypt/2.0.0

mattss:
This fixes quite a few issues with user profiles:
- Sets the default encryption to bcrypt by requiring relevant version of dx.membrane (resolves #779 )
- Ensures passwords are encrypted for users created via API
- Ensures all approved users are added to 'members' group (even when created via barcelonetta)


mattss:
Please note - dexterity.membrane changed some of its behaviour names, so this PR also provides an upgrade step to correct the names for existing deployments.

mattss:
@pilz This is the PR that also lets you change profile photos in barcelonetta

PR #822, opened 2015-10-07
Add image picker for the raptor editor.

jcbrand:
Create a new browser view @@panel-image-picker which shows a modal from which images can be picked to be inserted into the raptor editor.

Updated JS bundle with newest pat-raptor and patternslib.

Also ran `make diazo` to get newest static resources.

Updated HTML markup in document_view.html to look like the markup currently in the proto.

CC @pilz


jcbrand:
Please test this please

PR #832, opened 2015-10-09
Add support for apache virtual hosts in themeswitcher

mattss:
Adds support for looking up host from the Apache 'HTTP_X_FORWARDED_HOST' header

### DIFF
--- a/src/ploneintranet/themeswitcher/policy.py
+++ b/src/ploneintranet/themeswitcher/policy.py
@@ -137,10 +137,18 @@ class SwitchableThemingPolicy(DefaultPolicy):
 
     def getHostname(self):
         """Extract hostname from request."""
-        host = self.request.get('HTTP_HOST')
-        if not host:  # fallback is ''
-            host = self.request.get('SERVER_NAME')
-        return host.split(':')[0]  # ignore port
+        host = None
+        for key in [
+            'HTTP_X_FORWARDED_HOST',
+            'HTTP_HOST',
+            'SERVER_NAME',
+        ]:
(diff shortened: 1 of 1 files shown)


mattss:
Jenkins test this please.

src/ploneintranet/themeswitcher/policy.py
```diff
@@ -137,10 +137,18 @@ def getSwitcherSettings(self):
...
-        host = self.request.get('HTTP_HOST')
-        if not host:  # fallback is ''
-            host = self.request.get('SERVER_NAME')
-        return host.split(':')[0]  # ignore port
+        host = None
+        for key in [
+            'HTTP_X_FORWARDED_HOST',
+            'HTTP_HOST',
+            'SERVER_NAME',
+        ]:
```
mgrbyte, line 145: I'm surprised this doesn't fail pep8 checks (]: on its own line?)

ale-rt, line 145: It  is perfectly pep8 compatible! 
I like and support this coding style a lot.

It has several advantages, the first that come to my mind:
- if you list get longer you don't have to split it at a later stage
- if you modify it diff looks better
- you can rearrange the order of the list item with few key strokes

The same applies for tuple, sets, dicts, and so on...
I am using the same approach also on tal:define and tal:attributes :)

If I had to say something I would just complain about defining host as None instead of an empty string, I don't like variables to change type... but this is really a matter of taste :)

mattss:
Jenkins test this please.

Issue #835, opened 2015-10-14
Add a transform to ensure global-statusmessage is injected


ale-rt:
Add the moment the developer wants, most of the time, to add to the injection parameters the selectors for the global-statusmessage:

```
data-pat-inject="... && source: #global-statusmessage; target: #global-statusmessage; && ..."
```

I would like to make this automatic by registering a plone transform that adds this parameters to data-pat-inject (if not already present).
IMHO this will save us many headaches, given that we inject the status message all the time.

mattss:
Rather than a transform, I think there should be some kind of global settings for pat-inject, which allows us to make sure that both status message and exceptions (or just non-200 responses) are _always_ dealt with in a user-friendly manner, and avoid the endless spinner of death. /cc @jcbrand

PR #836, opened 2015-10-14
The insufficient privileges has a status message

### LINKED ISSUE #503: Insufficient Privileges page loaded via injection should show itself in the notifications
Currently, as it doesn't contain the selectors, the injection expects, the eternal spinner is shown. If the spinners show but a notification says "Insufficient Privileges", the user is at least informed why this happens.

This should be as easy as overriding the insufficient privileges view and add a container for the status messages.

ale-rt:
Fixes #503 

Anyway the message is shown if and only if the developer adds to data-pat-inject the correct parametes.
I have an idea on how to fix this (see #835).

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -23,3 +23,10 @@ parts +=
 
 [scripts]
 eggs = ${buildout:package-name} ${buildout:package-extras}
+
+[instance]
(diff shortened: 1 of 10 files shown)


ale-rt:
Superseeded by quaive/ploneintranet#205

PR #837, opened 2015-10-14
Change default entry in selects

### LINKED ISSUE #585: No value for selects
It's probably a bit friendlier/less technical to write 'Unknown' instead of 'No value' for selects that need a first option like this.

image

ale-rt:
The default from z3c.form is "No value" I used "Unknown as suggested"

Closes #585


ale-rt:
I did not write a robot test because grepping the code I did not find an obvious page where to test it.
Anyway I tried it with a custom form and it works nice

ale-rt:
Superseeded by quaive/ploneintranet#203

PR #841, opened 2015-10-15
Append the webstats_js view

### LINKED ISSUE #819: Support for analytics code
Plone supports an analytics code snippet, but I don't think that we currently support this in Plone Intranet. A simple win would be to update the diazo rules to bring it across from Plone.

@gyst Any thoughts on the priority of this? It's a pretty important feature for websites, but not so much for private intranets possibly.

ale-rt:
Fixes #819


ale-rt:
jenkins test this please

ale-rt:
Superseeded by quaive/ploneintranet#202

PR #843, opened 2015-10-15
Understand ajax_load parameter

ale-rt:
Since Plone 5 we handle ajax_load calls with two templates, I am shortcutting ajax_template to main_template.

Without this patch if you append ?ajax_load=1 to any ploneintranet URL you will have errors in the Plone logs.


ale-rt:
Superseeded by quaive/ploneintranet#204

Issue #629, opened 2015-07-23
power search


gyst:
For library tag faceting, I need to be able to query (tag:foo AND tag:bar). That's not possible with the current search implementation that hardcodes OR filters. This currently forces me to work around that by searching for SearchableText:foo+tag:bar which happens to give AND semantics close to what I need.

The OR logic is hardcoded in
https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/search/solr/utilities.py#L177

My proposal is to make the line quoted above smarter. For example by making it possible to,
instead of 

```
filters=dict(tags=['foo','bar'))
```

supply a 

```
filters=dict(tags=TERM('foo').OR(TERM('baz').AND(NOT('bar')))
```

I.e. I propose to create a DSL for chainable AND/OR/NOT/TERM and let the solr parser detect those.

This raises the question about how to handle ZCatalog degradation, which is a PITA anyway, see
https://github.com/ploneintranet/ploneintranet/commit/66ef08d6b4b2d972f66a5464deb573f213f943e7
Probably should be exposed as a capability and raise if not supported, letting the calling logic decide how to degrade itself on ZCatalog.

gyst:
http://docs.sympy.org/0.7.1/modules/logic.html

gyst:
An easier and more powerful alternative would be to support a .rawSolrQuery() method that allows you to compose the exact query in the calling application.

A related API enrichment would be to support getObject() on a result.

The combination of raw solr queries with result.getObject offers maximal power and flexibility in leveraging solr for application logic.

bfackland:
Maybe @mgrbyte and @mattss can help move this forwards.

mgrbyte:
With scorched, you can [arbitrarily combine](http://scorched.readthedocs.org/en/latest/query.html#complex-queries) `Q` objects;
I think this might be the best way to provide the desired behaviour.

I'm thinking we could/should permit:

``` python
  filters=Q('foo') | Q('bar') & ~Q('baz')
```

i.e switch on the filters parameter being a [query object](https://github.com/lugensa/scorched/blob/master/scorched/search.py#L24).

Edit: show combination of OR and AND and NOT

mattss:
+1 to that suggestion @mgrbyte

mattss:
ISearchResult items could potentially provide a 'geObject'-like behaviour by just doing an api.content.get on the existing 'path' attribute.

gyst:
[...]

PR #861, opened 2015-10-30
Support for complex solr queries (Implements request in #629)

mgrbyte:
@gyst Hopefully this meets your needs!
- Exposes the `execute` method, which can optionally be passed `secure=False` for unrestricted search
  - `execute` under ZCatalog implementation switches between `searchResults` and `unrestrictedSearchResults` for completeness. 
- Adds ability to pass a "complex" query object as a filter to the `query` method of `ploneintranet.search.ISiteSearch`
- Adds two solr-specific test cases to demonstrate usage
  - Re-factored connection into a property-like descriptor
  - Made `SiteSearchProtocol`'s doc-strings Sphinx compatible.

### DIFF
--- a/src/ploneintranet/search/base.py
+++ b/src/ploneintranet/search/base.py
@@ -7,6 +7,7 @@ import logging
 from plone import api
 from plone.api.validation import at_least_one_of
 from ploneintranet import api as pi_api
+from scorched.search import LuceneQuery
 from zope import globalrequest
 
 from .interfaces import ISearchResult
@@ -203,28 +204,29 @@ class SiteSearchProtocol:
 
     @abc.abstractmethod
     def _create_query_object(self, phrase):
-        """Return the query object to be executed from the given `phrase`."""
+        """Create a scorched query object for the given ``phrase``.
 
-    @abc.abstractmethod
-    def _apply_security(self, query):
-        """Apply a security filter to the query.
+        Apply any boost paramters based on plone.app.registry entries.
 
-        This should Plone's `allowedRolesAndUsers` index into account.
-        Return a copy of the query object with the security filter applied.
+        :return: The query object to be executed from the given ``phrase``.
         """
 
     @abc.abstractmethod
-    def _apply_filters(self, query, filters=None):
+    def _apply_filters(self, query, filters):
         """Build a filtered phrase query.
 
-        Return a query object that can be further filtered.
+        Ensure filters are a subset of those configured via plone.app.registry.
+
+        :return: A query object that can be further filtered.
         """
 
     @abc.abstractmethod
     def _apply_facets(self, query):
         """Apply parameters such that the query response will be faceted.
 
-        Return a copy of the modified `query`.
+        Ensure facets are a subset of those configured via plone.app.registry.
+
+        :return: A copy of the modified `query`.
         """
 
     @abc.abstractmethod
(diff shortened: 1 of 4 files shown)


gyst:
@mgrbyte looks good. However, I've tried to apply this to change the library tag logic that was the origin of this feature request, but I don't see a clean way to do that.

If you look at:
https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/library/browser/views.py#L182

What this wants to do is basically wrap an existing query specification (a dict, not an IQuery or a LuceneQuery) with an extra condition: AND tag=self.request_tag. The way you've implemented makes it easy to chain Lucene queries but it's not clear to me how I can chain filter dicts.

How can L182 be changed to best benefit from the refactored search you've implemented?

gyst:
Additionally, a `sort` parameter on `ISiteSearch.query()` would be very welcome to solve the likes of #862.

mgrbyte:
@gyst Its not possible to directly "chain filter dicts", you'll need to convert them to `LuceneQuery` objects first. 

An example of how this done is in the implmentation of the SiteSearch utllity's _apply_filters method.

Examples:

``` python
ss = getUtility(ISiteSearch)
```

Convert dict of parameters into an "AND" query

``` python
Q = ss.connection.Q
params = dict(path_parents='/library', tags=['foo'], portal_type='Image')
lucene_query = Q(**params)
print lucene_query
```

Convert dict of parameters into an "OR" query

``` python
Q = ss.connection.Q
params = dict(path_parents='/library', tags=['foo'], portal_type='Image')
lucene_query = Q()
for (key, val) in params.items():
    lucene_query |= Q(**{key: val})
print lucene_query
```

Applying the lucene query to perform a search:

``` python
# no phrase is same as '*:*'
query = ss.connection.query()
query = ss.filter(lucene_query)
# asc order
query = query.sort_by('portal_type')
# dsc order
query = query.sort_by('~portal_type')
ss.execute(query)
```

mgrbyte:
@gyst Please see the tests of [using site search](https://github.com/ploneintranet/ploneintranet/blob/power-search/src/ploneintranet/search/solr/tests/test_utilities.py#L53-L59) and [using the scorched API directly, via the site search connection](https://github.com/ploneintranet/ploneintranet/blob/power-search/src/ploneintranet/search/solr/tests/test_utilities.py#L61-L68)

mgrbyte:
Specifically with L182, I think you want (assuming you have the power-search branch):

``` python
Q = self.sitesearch.Q
return self.sitesearch.query(self.request_tag, filters=Q(filters), **kwargs)
```

If I read/infer the code and comments correctly, you could also just do, e.g: 

``` python
query = ss.connection.query(ss.Q('Needle') & ss.Q(tags='foo') & ss.Q(portal_type='Image')
# see the generated query
print query.query_obj
print query.options()
ss.execute(query)
```

gyst:
Thanks for your feedback @mgrbyte . I did read the tests :-) but your comments help clarify. I'll try and do some work on `power-search` later to factor these features into the API, will ask you to review.

gyst:
@mgrbyte I can't get this to work. I need a phraseless query to start with, and then to AND any filters. I've tried this on branch `power-search-library` see commit
https://github.com/ploneintranet/ploneintranet/commit/c72bb026512897e8727fda9020d7d49f30a9a78e

But I get empty tag views on http://localhost:8080/Plone/library/tag. This is because the searchresponse has no tag facets.

To reproduce:
- load genericsetup ploneintranet.suite:testing
- load that profile a second time -> this gives you a full library
- hit http://localhost:8080/Plone/library/tag

mattss:
@gyst I've done some work in a branch of your branch ;) here:

https://github.com/ploneintranet/ploneintranet/commits/power-search-library-redux

There are two commits - the first one is a small change that fixes the issue you were seeing. The faceting line was not re-assigning the variable properly.

The second commit you may disagree with - but I've tried to simplify things by using explicit keyword arguments and dealing with each on a case by case basis to build up the lucene query.

PR #866, opened 2015-11-03
pass a valid id to object creation. Otherwise p.api will first create…

pilz:
… an object with a random int id, do a partial commit and then rename it. This is stupid because all our async jobs will get started already for the temporary object. The workaround is easy and even documented by maurits inline in p.api.


mattss:
I looked into this a bit myself. It seems like a change that should go into plone.api itself.

ale-rt:
Already merged in quaive

PR #875, opened 2015-11-11
Refs #12692 Case: treat a WF state not in the MMap as a future state

cillianderoiste:
To allow extra workflow states which aren't used by the MetroMap
e.g. on-hold or rejected, we can treat these states the same as future
states in the MetroMap and set the WF state of currently Open
tasks to Planned, so that they don't appear on the Dashboard.

(cherry picked from commit 173235f918b6cd2029276f9e4ea1b5d40743f8f4)

### DIFF
--- a/src/ploneintranet/todo/content/todo.py
+++ b/src/ploneintranet/todo/content/todo.py
@@ -38,7 +38,14 @@ class Todo(Item):
             if milestone:
                 case_state = wft.getInfoFor(workspace, 'review_state')
                 mm = IMetroMap(workspace).metromap_sequence.keys()
-                future = mm.index(milestone) > mm.index(case_state)
+                # A case could be set to a state which isn't included in the
+                # metromap e.g. on-hold, rejected. If that happens we can treat
+                # the case_state as a future state and set all open tasks to
+                # planned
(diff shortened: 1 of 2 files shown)


src/ploneintranet/todo/content/todo.py
```diff
@@ -38,7 +38,14 @@ def reopen(self):
...
                 mm = IMetroMap(workspace).metromap_sequence.keys()
-                future = mm.index(milestone) > mm.index(case_state)
+                # A case could be set to a state which isn't included in the
+                # metromap e.g. on-hold, rejected. If that happens we can treat
+                # the case_state as a future state and set all open tasks to
+                # planned
+                future = (
+                    case_state not in mm or
+                    mm.index(milestone) > mm.index(case_state)
+                )
```
mgrbyte, line 48: @cillianderoiste Could `index()` potentially raise `ValueError` here? (i.e Are `milestone` and `case_state` guaranteed to be keys of the metromap_sequence?)

cillianderoiste, line 48: Thanks @mgrbyte, you're right, I'll fix it.

Issue #867, opened 2015-11-04
ploneintranet.search - unicode vs bytes issue


mgrbyte:
@gyst, @mattss 
Currently the tests for ploneintranet.search ambiguous with regard to its handling of strings (unicode vs bytes)

The `query` method is tested with both bytes and unicode inputs, and currently assets (by accident more than anything else!) that it gets bytes back.  This is due to using the 'DublicCore' accessors such as `Title()`. An example of this is [here](https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/search/tests/base.py#L141-L150)

I believe that search search should accept text input only (strings of type unicode) and the ISearchResponse object should present strings as text only (type unicode).

Currently what it does is to present things like title as unicode encoded bytes (strings of type str)

See

gyst:
I've wasted hours of my life finding out the hard way, that the ploneintranet.search implementations for solr and zcatalog behave differently in this regard. I expect the zcatalog implementation to be a nuisance when doing this, and I've come to view that implementation as a dead weight anyway, so you can probably make life easy for yourself by ripping out zcatalog completely and moving all tests to the solr layer, when forcing unicode which is +100.

PR #876, opened 2015-11-12
Use ISiteSearch (i.e. solr if available) in sidebar search

reinhardt:
This is another prerequisite for dropping SearchableText from the portal_catalog, s.a. #872.

Includes some changes to SearchResult, making it more like a catalog brain.

### DIFF
--- a/buildout.d/solr.cfg
+++ b/buildout.d/solr.cfg
@@ -38,6 +38,7 @@ index =
     name:Description            type:text copyfield:default copyfield:spelling stored:true
     name:SearchableText         type:text copyfield:default copyfield:spelling stored:false
     name:tags                   type:string copyfield:default stored:true multivalued:true
+    name:Subject                type:string stored:true indexed:false multivalued:true
     name:Title                  type:text copyfield:default copyfield:spelling stored:true
     name:UID                    type:string stored:true required:true
     name:allowedRolesAndUsers   type:string stored:false multivalued:true
@@ -46,6 +47,7 @@ index =
     name:expires                type:date stored:true
     name:email                  type:string stored:true
     name:friendly_type_name     type:string stored:true indexed:true
+    name:getId                  type:string stored:true indexed:false
     name:has_thumbs             type:boolean indexed:true
     name:modified               type:date stored:true
     name:path_depth             type:integer indexed:true stored:false
--- a/src/ploneintranet/workspace/browser/tiles/sidebar.py
+++ b/src/ploneintranet/workspace/browser/tiles/sidebar.py
@@ -17,12 +17,14 @@ from plone import api
 from plone.app.contenttypes.interfaces import IEvent
 from plone.app.event.base import localized_now
 from plone.i18n.normalizer import idnormalizer
+from ploneintranet.search.interfaces import ISiteSearch
 from ploneintranet.todo.utils import update_task_status
 from Products.CMFPlone.utils import safe_unicode
 from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
 from Products.statusmessages.interfaces import IStatusMessage
 from zope.component import getAdapter
 from zope.component import getMultiAdapter
+from zope.component import getUtility
 from zope.event import notify
 from zope.lifecycleevent import ObjectModifiedEvent
 from zope.publisher.browser import BrowserView
@@ -452,7 +454,7 @@ class Sidebar(BaseTile):
                 url=url,
                 creator=api.user.get(username=r['Creator']),
                 modified=r['modified'],
-                subject=r['Subject'],
+                subject=r.get('Subject', ()),
                 UID=r['UID'],
                 path=r.getPath()
             ))
@@ -482,7 +484,6 @@ class Sidebar(BaseTile):
         It returns the items based on the selected grouping (and later may
         take sorting, archiving etc into account)
         """
-        catalog = api.portal.get_tool("portal_catalog")
         current_path = '/'.join(self.context.getPhysicalPath())
         sidebar_search = self.request.get('sidebar-search', None)
         query = {}
@@ -502,11 +503,14 @@ class Sidebar(BaseTile):
         if sidebar_search:
             # User has typed a search query
 
+            sitesearch = getUtility(ISiteSearch)
+            query['path'] = current_path
             # XXX plone only allows * as postfix.
             # With solr we might want to do real substr
-            query['SearchableText'] = '%s*' % sidebar_search
-            query['path'] = current_path
-            results = self._extract_attrs(catalog.searchResults(query))
+            response = sitesearch.query(phrase='%s*' % sidebar_search,
+                                        filters=query,
+                                        step=99999)
+            results = self._extract_attrs(response)
 
         elif self.request.get('groupname'):
             # User has clicked on a specific group header
--- a/src/ploneintranet/search/base.py
+++ b/src/ploneintranet/search/base.py
@@ -90,6 +90,7 @@ class SearchResult(object):
         self.portal_type = context['portal_type']
         self.contact_email = context.get('email')
         self.contact_telephone = context.get('telephone')
+        self.mimetype = context.get('mimetype', None)
         if context['has_thumbs']:  # indexer in docconv
             # can occur in workspaces AND library
             if self.portal_type in ('Image', 'Document', 'News Item'):
@@ -135,6 +136,12 @@ class SearchResult(object):
     def path(self):
         """Return the path URI to the object represented."""
 
+    def getPath(self):
+        return self.path
+
+    def getId(self):
+        return self.context['getId']
+
     @property
     def url(self):
         """Generate the absolute URL for the indexed document.
@@ -149,6 +156,9 @@ class SearchResult(object):
             url = '{}/view'.format(url)
         return url
 
+    def getURL(self):
+        return self.url
+
     @property
     def preview_image_url(self):
         """
@@ -158,6 +168,15 @@ class SearchResult(object):
         :rtype: str
         """
         return self._path_to_url(self.preview_image_path)
+
+    def __getitem__(self, key):
+        return self.context[key]
+
+    def get(self, key, default=None):
+        try:
+            return self[key]
+        except KeyError:
+            return default
 
 
 class SearchResponse(collections.Iterable):


buildout.d/solr.cfg
```diff
@@ -38,6 +38,7 @@ index =
     name:Description            type:text copyfield:default copyfield:spelling stored:true
     name:SearchableText         type:text copyfield:default copyfield:spelling stored:false
     name:tags                   type:string copyfield:default stored:true multivalued:true
+    name:Subject                type:string stored:true indexed:false multivalued:true
```
mattss, line 41: We shouldn't need both tags and Subject. The ploneintranet.search code pulls the tag values out of plone's 'Subject' field already.

src/ploneintranet/workspace/browser/tiles/sidebar.py
```diff
@@ -452,7 +454,7 @@ def _extract_attrs(self, catalog_results):
                 url=url,
                 creator=api.user.get(username=r['Creator']),
                 modified=r['modified'],
-                subject=r['Subject'],
+                subject=r.get('Subject', ()),
                 UID=r['UID'],
                 path=r.getPath()
             ))
```
mattss, line 460: I'd rather change this code to read the variables that the search api already provides, rather than adding extra methods to the api.

mattss:
The search API is pretty well defined, and already provides most of what you need from a brain, but in a more generic format. I would much rather we just changed any calling code to use the api properties provided.

reinhardt:
Ah, sorry, I didn't notice that there was already a "tags" index and that it indexes the"Subject" field.
In the sidebar it would be great if we could reuse the method that extracts information without checking whether we have a catalog brain or a search result, e.g. whether we need to look for "Subject" or for "tags". I could change the SearchResult so that the tags are available under both keys.

mattss:
The whole Subject/tags thing in ploneintranet.search is kind of nuts, largely because 'subject' always seemed an unhelpful term (I think it comes from dublin core?). My preferred option would actually be just to register a 'tags' indexer that converts from Subject -> tags at the catalog level, and then we can just use tags everywhere.

mgrbyte:
@mattss I think your comment regarding tags is valid.
Furthermore, I think the indexing of values obtained via `ICMFDublinCore` should not be done at all.
I think this needs addressing in the context of the [unicode vs bytes](#867), which applies to search but also code-base in general.

Currently, we index the following things, using values obtained (I believe, need to audit) via `zope.dublincore.interfaces.ICMFDublinCore`:
    \* Title
    \* Description

I think we should avoid these, because `plone.dexterity.content.DexterityContent` transforms these values to bytes, where we should be using unicode everywhere (except at i/o boundaries).

Instead, we should use the dx attributes `title` and `description`.

What do people think?
/cc @gyst @pilz

gyst:
In general I'm of the persuasion that behavior managed attributes should be considered as private and only be accessed via the accessors provided by the behavior, i.e. by adapting the context to the behavior.

In the case of DublinCore, like Subject (= tags indeed), Title and Description everybody should be aware that the relevant implementation is provided by https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/network/behaviors/metadata.py. The reason for that blanket fork is documented here: https://community.plone.org/t/how-to-change-existing-dexterity-types-and-behaviors/219.

As it is, the indexing of title and description is inherited from upstream /somewhere/ so let's not go there and just keep indexing the attributes as-is -- we don't need to override the indexers. We can override the setter and getter if we want to change the unicode behavior there in our own IDublinCore override.

mgrbyte:
@gyst  Thank you for your detailed and comprehensive response.

This makes me :sob:

Instinctively, after reading this and the links you provided, I wanted to have a go at a patch that would not involve having to fork... then I thought you probably did all these things already, and I would be wasting my time.

for all that it's worth, this disaster is as a direct result of the use of `sys._getfame` (in this case in `plone.supermodel` (via `plone.autoform.directives`).
Other culprits are `z3c.form` and `plone.directives.form` (and `zope.interface` itself, but that one is ok (I believe) due to the general usage pattern - with one or two caveats). 

In general, anywhere where we see someone use `sys._getframe` in library/framework code that's intended for _extensibility_: :facepunch: :facepunch: :facepunch: 

Your suggestion to to handle the unicode vs bytes issue in the dublin core override is probably the best that can be done, given the woe mentioned above.

I think the overriding of DublinCore should be given more prominence;
move it out of pi.network to it's own package into it's own new sub-package `ploneintranet.dublincore`.

I was not aware that this was being done (I wonder how many were?), 
furthermore, I was not aware that the _default_ dublincore behaviors are replaced by `ploneintranet.network`'s instllation via its setuphandlers .

I think this change will help in the future, especially if the DC implementation further deviates from  what zope/plone's DublinCore does by default.

Out of curiosity, (if you'd be so kind!) I'd like to understand the design rationale/necessity of employing the DublinCore `Subject` attribute/acessor for tags; 
e.g Why a new behavior (e.g ITaggable) which adds a `tags` property was not introduced instead?

Thanks.

gyst:
On 18-11-15 23:30, Matt Russell wrote:

> @gyst https://github.com/gyst Thank you for your detailed and
> comprehensive response.
> 
> This makes me :sob:

I'm sorry to hear that, that was never my intention. ;-)

> Instinctively, after reading this and the links you provided, I wanted
> to have a go at a patch that would not involve having to fork... then I
> thought you probably did all these things already, and I would be
> wasting my time.

Yeah, been there, done that, wasted a few days in another project. All I
wanted to to was add an extra field in an existing fieldset, and I ended
up with the horror of a full DublinCore fork, when all the more subtle
and smaller interventions I tried kept failing.

Asko confirmed the necessity of that in the community thread I
referenced before.

> for all that it's worth, this disaster is as a direct result of the use
> of |sys._getfame| (in this case in |plone.supermodel| (via
> |plone.autoform.directives|).
> Other culprits are |z3c.form| and |plone.directives.form| (and
> |zope.interface| itself, but that one is ok (I believe) due to the
> general usage pattern - with one or two caveats).

That's beyond my ken, please enlighten me. I've come to realize that
anywhere in the Plone stack, except behaviors, one can use the ZCA to
inject a polymorphism override. You would expect to be able to::

  from plone.whatever import IDublinCore, DublinCore

  class IMyDublinCore(IDublinCore):
    pass

  class MyDublinCore(DublinCore):
    implements(IMyDublinCore)

Hook that up in ZCML and presto, you've got yourself a behavior override
that delegates anything you have not tweaked up to the parent, because
IMyDublinCore is a /more specific interface/ than IDublinCore and steals
all the IDublinCore clients, effectively.

For behaviors, however, these appear not to be looked up based in
interface resolution order, but are looked up hardcoded by dotted name.
And if you put your own behavior in your own module, it has another
dotted name innit and it's a different behavior. Which still doesn't
explain why the subclassing doesn't work at all, but that is probably
the supermodel's doing? Would be happy to learn how sys._getframe
contributes to this mess.

> I think the overriding of DublinCore should be given more prominence;
> move it out of pi.network to it's own package into it's own new
> sub-package |ploneintranet.dublincore|.

I get your reasoning but at this stage do not agree, because the only
active consumer of the override is ploneintranet.network itself and
there's a lot of tight coupling with the ploneintranet.network data
structures, see below.

As soon as we get more changes in our IDublinCore unrelated to
ploneintranet.network, I agree it should be factored out of network.

> I was not aware that this was being done (I wonder how many were?),
> furthermore, I was not aware that the /default/ dublincore behaviors are
> replaced by |ploneintranet.network|'s instllation via its setuphandlers .
> 
> I think this change will help in the future, especially if the DC
> implementation further deviates from what zope/plone's DublinCore does
> by default.
> 
> Out of curiosity, (if you'd be so kind!) I'd like to understand the
> design rationale/necessity of employing the DublinCore |Subject|
> attribute/acessor for tags;

The rationale for this is, that:
- I wanted to provide a personalized tagging system, which keeps track
  of all the tags you yourself used on which documents
- So we can provide better (personalized) tagging suggestions, based on
  your historical tag usage
- While maintaining backward-compatibility with the "global" tag set of
  documents (which is indexed by solr)

Having this in place is part of a long-term strategy to work towards a
machine learning system that provides tag suggestions and manages it's
own tag set logged in as a bot user.

The behavior override is here:

https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/network/behaviors/metadata.py#L355

You can see that the getter does nothing special, just the subject
attribute.

But the setter is a completely different piece of cake.
361 is a unicode fix that ought to make you happy
362 is the backward compatibility setter to make the getter work
Anything beyond that is the personalized tagging system, that basically
stores a (user, tag, object) triplet in the graph store and indexes it
internally. That happens here:

https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/network/graph.py#L262

This graph database came into being as tracking
1. <user> follows <user>
2. <user> endorses <user> with <skill>
3. <user> tags <object type> <object> with <tag>

As you can see, 2 extends 1, and 3 is a generalization of 2.

So we have two subject (tag) storages: the context.subject attribute,
and the ploneintranet.network graph database with personalized tags.

Tag suggestions are pulled in via a widget override
https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/network/browser/vocabulary.py

Which pulls in a tag vocabulary override
https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/network/vocabularies.py

If you feel like doing some more weeping, the full call flow is
documented here:

https://github.com/ploneintranet/ploneintranet/blob/master/src/ploneintranet/network/behaviors/metadata.py#L122

> e.g Why a new behavior (e.g ITaggable) which adds a |tags| property was
> not introduced instead?

Maintaining maximal backward compatibility. The solution outlined above:
1. Introduces personalized tagging without requiring changes in the solr
and zcatalog indexes (and their consumers!!!)
2. Can be run from a Barceloneta theme without ploneintranet.theme
overrides, which is a hard requirement since it should also work in
library editing which is done via Barceloneta
3. Can be installed and uninstalled without requiring database
migrations, since it maintains the existing upstream context.subject
attribute intact at all times.

Another way of saying this, is that changing context.subject to
context.tags would require massive changes throughout not only our own
code base but also many upstream components, would create a barrier to
re-using existing Plone add-ons depending on 'subject' within Quaive,
and would make it impossible to re-use the personalized tagging backend
outside of Quaive.

> Thanks.

You're welcome!

## 

```

```

mgrbyte:
@gyst Thanks again for a wonderfully detailed reply.

I will need to re-read this (to take it all in properly) later.

re: sys._getrame - I'll try and find some references +/ find time to come up with an example if I can craft one.

mgrbyte:
@gyst fwiw, re: sys._getframe
Here's a couple of references:

In context of decorators:
http://stackoverflow.com/questions/22541526/decorators-influencing-sys-getframe1

Furthermore, Guido's sentiment in 2004 (which am almost certain still holds true)
http://article.gmane.org/gmane.comp.python.devel/58837

A call to: `sys._getfame(1)` will always refer to the module in which its called.

I think I'll end this here now; we should perhaps open a new issue if there's more to discuss (I'm sure there will be)

PR #885, opened 2015-11-18
Fix workspace tests

mattss:
Pull request #872 broke the workspace tests. This is an attempt to fix them.

/cc @pilz


pilz:
Thank you!

PR #889, opened 2015-11-26
Sanitize HTML input using htmllaundry for RichText

pysailor:
When we save a Richtext field, apply HTML sanitizing using htmllaundry

Reason: strip any tags that got pasted in accidentally that are not wanted (in case the editor has not stripped them).
This applies e.g. to `<span>`. The reason we define our own DocumentCleaner is that we
want to allow h1 and h2 tags, wich the default DocumentCleaner does not allow and strip
other tags like `<blockquote>` that are part of the default set.
We also make sure that no text is present outside of a block-level element.
A basic unit test is also added - it is far from exhaustive but should alert us in
case sanitizig breaks.


pysailor:
Refactored and squashed commits. Maybe @ale-rt or @mattss could you please have a look and tell me if you think this makes sense.

gyst:
For such a crucial component, the test coverage appears to be weak. I think `sanitize_html()` deserves some extra unittests that exercise the nested `body` logic, and generally probe resilience against some really malformed Word html.

ale-rt:
@pysailor I am closing this one because it was moved here quaive/ploneintranet#173
Correct me if I am wrong

pysailor:
@ale-rt oh yeah, that was my ancient first attempt, long outdated :)

